PR #1, opened 2022-01-24
Add `py-feed-animals` task

IvanRamyk:


tests/test_main.py
```diff
@@ -1,6 +1,182 @@
...
+
+
+@pytest.mark.parametrize(
+    "class_name,method",
+    [
+        ("Cat", "catch_mouse"),
+        ("Dog", "bring_slippers"),
+    ],
+)
+def test_only_one_method_should_be_declared_in_each_of_children_classes(
```
danylott, line 39: Good job!
You've automated even this check with tests!
(In GitHub tasks sometimes we can check just with eyes, but this is really good to check it this way :)

danylott (approved):
LGTM

tests/test_main.py
IvanRamyk: I just copied it with new params from code editor task :)
From pytest.mark.parametrize with love)

MasterpieceElbow (approved):

PR #5, opened 2022-01-25
'Solution'

liuda-chuzdiuk:


app/main.py
```diff
@@ -10,6 +10,7 @@ def print_name(self):
     def feed(self):
         print(f"Eating {self.appetite} food points...")
         if self.is_hungry is True:
```
phantom-terminal, line 12: Maybe you should check to see if the animal is hungry first and then feed it

phantom-terminal (changes requested):

app/main.py
liuda-chuzdiuk: Good point!

phantom-terminal (approved):

MasterpieceElbow (approved):

PR #6, opened 2022-01-25
'Solution'

cloudprogramm:


app/main.py
```diff
@@ -1 +1,45 @@
...
+    def catch_mouse():
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry=True):
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
+        print()
```
IgorOlifirenko, line 37: Remove this print)

IgorOlifirenko (changes requested):

IgorOlifirenko (approved):

app/main.py
```diff
@@ -1 +1,44 @@
...
+        else:
+            return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry=True):
+
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
MasterpieceElbow, line 25: Your parent's init can set `self.is_hungry`, why do you duplicate yourself?

app/main.py
```diff
@@ -1 +1,44 @@
...
+    @staticmethod
+    def catch_mouse():
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry=True):
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
```
MasterpieceElbow, line 36: The same as Cat

MasterpieceElbow (changes requested):

app/main.py
cloudprogramm: How I can use non positional argument after positinal argument?

MasterpieceElbow (approved):

PR #13, opened 2022-01-25
'Solution'

denyslipin:


app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def feed(self):
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
```
alexandra-halytska, line 18: I had the same problem, as Oleksii Proshchenko  said: "Why do you duplicate code from your parent Animal? In Animal you are already set self.name and self.is_hungry. Always pass to super() attrs that are the same with parent and after add attrs that are different"

app/main.py
```diff
@@ -1 +1,41 @@
...
+        super().__init__(name, appetite=3)
+        self.name = name
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def catch_mouse():
+        print("The hunt began!")
+
+
+class Dog(Animal):
```
alexandra-halytska, line 29: the same

alexandra-halytska (approved):

MasterpieceElbow (approved):

PR #18, opened 2022-01-26
Solution

roland813:


Oleksandr563487953 (changes requested):
Function feed_animals() can be much more simple, just in one line. You can use the feed() method  in 'for' loop to obtain appetite attribute if animal is hungry, and get the sum of appetites.

roland813:
Chenged. And use list comprehension. Thanks for the advice

Oleksandr563487953 (approved):

app/main.py
```diff
@@ -1 +1,47 @@
...
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name, is_hungry=True):
+        self.name = name
+        self.is_hungry = is_hungry
```
MasterpieceElbow, line 24: Why do you assign this, it you pass `name` and `is_hungry` to super().__ init__?

app/main.py
```diff
@@ -1 +1,47 @@
...
+    @staticmethod
+    def catch_mouse():
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name, is_hungry=True):
+        self.name = name
+        self.is_hungry = is_hungry
```
MasterpieceElbow, line 36: The as Cat.

MasterpieceElbow (changes requested):

Zikkolion (approved):

MasterpieceElbow (approved):

PR #19, opened 2022-01-26
'Solution'

Oleksandr563487953:


app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
nikon4uk, line 15: not necessary use else

app/main.py
```diff
@@ -1 +1,44 @@
...
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name, appetite=3, is_hungry=True):
+        super().__init__(name, appetite, is_hungry)
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
```
nikon4uk, line 24: you passed the same argumenst into father constuctor with function super, shoudn`t reassign them below. If you want set default values you can do it when call method super().__init__(variable=default_value)

app/main.py
```diff
@@ -1 +1,44 @@
...
+    def catch_mouse():
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name, appetite=7, is_hungry=True):
+        super().__init__(name, appetite, is_hungry)
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
```
nikon4uk, line 36: same case

nikon4uk (changes requested):

app/main.py
Oleksandr563487953: All changes provided. You helped me to understand the super() method, thanks!

nikon4uk (approved):
good job!

MasterpieceElbow (approved):

PR #21, opened 2022-01-26
'Solution'

Zikkolion:


app/main.py
```diff
@@ -1 +1,36 @@
...
+class Animal:
+    def __init__(self, name: str, appetite: int, is_hungry: bool = True):
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
```
8Men5Camels, line 10: here you need to add a check - whether the animal is hungry. And then do all the manipulations below or return 0

app/main.py
```diff
@@ -13,6 +13,8 @@ def feed(self):
             self.is_hungry = False
             food_points, self.appetite = self.appetite, 0
```
8Men5Camels, line 14: we can't change  - "appetite", this value must remain unchanged.
You need to return the number of units eaten depending on the status of "is_hungry"

IvanRamyk (approved):

PR #26, opened 2022-05-19
Solution

YuriiKindrat:


DmytroSvirsa (approved):

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    @staticmethod
+    def bring_slippers():
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list):
+    return sum([animal.appetite
+                for animal in list_of_animals
+                if animal.is_hungry])
```
Magus6784, line 42: your poor animals are still hungry. We need use "feed" because in feed we have self.is_hungry = False

Magus6784 (changes requested):

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    @staticmethod
+    def bring_slippers():
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list):
+    return sum([animal.feed()
+                for animal in list_of_animals
+                if animal.is_hungry])
```
Magus6784, line 42: You can check Slack, mentor says we dont need feed.
So... Sorry)

Magus6784 (changes requested):

PR #30, opened 2022-05-23
Solution

darten360:


app/main.py
```diff
@@ -1 +1,39 @@
...
+    def __init__(self, name: str, is_hungry: bool = True):
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers():
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list):
+    return sum(animal.appetite for animal in animals if animal.is_hungry)
```
DmytroSvirsa, line 40: For what purpose do you have a feed method? You do not feed animals, you count how much they need and steal food from animals. But, it works correctly.

DmytroSvirsa (approved):

PR #38, opened 2022-05-23
Solution

vasylhnatiuk:


app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Flashmobber, line 16: This "else" is not needed here

app/main.py
```diff
@@ -1 +1,42 @@
...
+            return 0
+
+
+def feed_animals(animals: list):
+    return sum([
+        animal.appetite for animal in animals
+        if animal.is_hungry is True
+    ])
```
Flashmobber, line 24: It's a bad practice. Move this function after the class description please.

Flashmobber (changes requested):

PR #40, opened 2022-05-23
Solution

NOSSUDAK:


app/main.py
```diff
@@ -1 +1,46 @@
...
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        print(f"Eating {self.appetite} food points...")
+        was_hungry = self.is_hungry
+        self.is_hungry = False
+        if was_hungry:
+            return self.appetite
+        return 0
```
Magus6784, line 21: I'm not sure if this is a problem or not, but I think it's a bit ugly
Maybe better don't use extra variable
`        if self.is_hungry:
            self.is_hungry = False
            return self.appetite`

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True):
+        appetite = 3
+        super().__init__(name, appetite, is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
```
Magus6784, line 30: This is     @staticmethod, so rewrite with him

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True):
+        appetite = 7
+        super().__init__(name, appetite, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
```
Magus6784, line 39: same with $29 - $30

Magus6784 (changes requested):

app/main.py
NOSSUDAK: Should we feed animal if it's not hungry? Nothing said about that in readme

app/main.py
NOSSUDAK: You make a good point

Magus6784 (approved):

MasterpieceElbow (approved):

lenasharabura (approved):

MasterpieceElbow (approved):

PR #43, opened 2022-05-23
Solution

Hectorovich:


app/main.py
```diff
@@ -1 +1,41 @@
...
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name, is_hungry=True):
+        super(Cat, self).__init__(name, 3, is_hungry)
```
katevasilenko, line 20: I'm not sure you need to fix it, but if could to not write in braskets after super, so why to write

katevasilenko (changes requested):

PR #44, opened 2022-05-23
'Solution'

YuraBeseda:


app/main.py
```diff
@@ -1 +1,38 @@
-#  write your code here
+class Animal:
+    def __init__(self, name, appetite, is_hungry=True):
```
Serhii-Leonenko, line 2: It's will be better if you use type annotation)

app/main.py
```diff
@@ -1 +1,38 @@
...
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
Serhii-Leonenko, line 16: You can remove `else` here, because you have used `return self.appetite`  after `if` .

Serhii-Leonenko (changes requested):

MasterpieceElbow (approved):

PR #46, opened 2022-05-23
py-feed-animals solution

Y-Havryliv:


app/main.py
```diff
@@ -1 +1,54 @@
...
+    ):
+        super(Dog, self).__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers():
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list):
+    return sum(Animal.feed(animal) for animal in animals)
```
Flashmobber, line 54: I think here you are making a direct call to the class method by passing it an instance of the class created in the test before. As I understand the requirement of the task, here you need to collect data on the appetite of animals, and all the rest is maked by the test.

Flashmobber (changes requested):

Flashmobber (approved):

PR #47, opened 2022-05-23
'feed-animals'

CheshkoSergey:


app/main.py
```diff
@@ -1 +1,43 @@
...
+            self.is_hungry = False
+            points = self.appetite
+            self.appetite = 0
+            return points
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True):
+        super(Cat, self).__init__(name, 3, is_hungry)
```
daria-niklaevna, line 24: It is not clear what "3" is, it is better to indicate the name of the variable to which we assign this value.

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    @staticmethod
+    def catch_mouse():
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True):
+        super(Dog, self).__init__(name, 7, is_hungry)
```
daria-niklaevna, line 34: Same problem as in line 24.

app/main.py
```diff
@@ -1 +1,43 @@
-#  write your code here
+class Animal:
+    animals = []
```
daria-niklaevna, line 2: We do not use this variable, we can remove it.

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        if self.is_hungry is True:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            points = self.appetite
+            self.appetite = 0
```
MasterpieceElbow, line 16: Why do you change `self.appetite`?

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #50, opened 2022-05-23
feed-animals Solution

mens0:


app/main.py
```diff
@@ -1 +1,38 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name, is_hungry=True):
+        super(Cat, self).__init__(name, 3, is_hungry)
```
daria-niklaevna, line 21: It is not clear what "3" is, it is better to indicate the name of the variable to which we assign this value.

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+    @staticmethod
+    def catch_mouse():
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name, is_hungry=True):
+        super(Dog, self).__init__(name, 7, is_hungry)
```
daria-niklaevna, line 30: Same problem as in line 21.

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
MasterpieceElbow, line 15: You don't have to use the `else` block if you have a `return` statement in the `if` block. Just `return 0`

MasterpieceElbow (changes requested):

PR #51, opened 2022-05-23
Katya's Solution

katevasilenko:


app/main.py
```diff
@@ -1 +1,43 @@
...
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        print(f"Eating {self.appetite} food points...")
+        if self.is_hungry is True:
```
RomaTsybulia, line 13: "self.is_hungry" is  true by default. You don't need additional check.

app/main.py
```diff
@@ -1 +1,43 @@
...
+    def __init__(self, name: str, is_hungry: bool = True):
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers():
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list):
+    food = 0
```
RomaTsybulia, line 40: you can use list comprehension, but it isn't necesary

RomaTsybulia (changes requested):

RomaTsybulia (approved):

MasterpieceElbow (approved):

PR #52, opened 2022-05-23
Solution

Chardona:


app/main.py
```diff
@@ -1 +1,39 @@
...
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, appetite: int = 3, is_hungry: bool = True):
```
4ndyua, line 20: It should have a constructor with two arguments:

name - the name of a cat
is_hungry - with True default value

app/main.py
```diff
@@ -1 +1,39 @@
...
+        super().__init__(name, appetite, is_hungry)
+
+    def catch_mouse(self):
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, appetite: int = 7, is_hungry: bool = True):
```
4ndyua, line 28: The last class you should implement is a Dog class. Its constructor should have two arguments:

name - the name of a dog
is_hungry - with True default value

4ndyua (changes requested):
Your constructors have 3 arguments instead of 2

4ndyua (approved):

app/main.py
```diff
@@ -1 +1,39 @@
...
+
+    def print_name(self):
+        print(f"Hello, I\'m {self.name}")
+
+    def feed(self):
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
MasterpieceElbow, line 15: You don’t need `else` if you have a `return` statement in `if`.

MasterpieceElbow (changes requested):

app/main.py
Chardona: I thought it's more for visual understanding of code, but I've changed.

MasterpieceElbow (approved):

PR #53, opened 2022-05-23
Solution

4ndyua:


app/main.py
```diff
@@ -1 +1,43 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name, is_hungry=True):
+        super().__init__(name, 3, is_hungry)
```
Chardona, line 22: 3 isn't obvious argument. Suggest changing into keyword argument.

Chardona (changes requested):
Non-obvious 3 and 7.

Chardona (approved):
Not sure why u've added is_hungry, but code seems okay.

PR #54, opened 2022-05-23
solution

BohdanKryven:


app/main.py
```diff
@@ -1 +1,44 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True):
+        super().__init__(name=name, appetite=3, is_hungry=is_hungry)
```
YuriiKindrat, line 21: you can just write here super().__init__(name, 3, is_hungry) 
and same on $30

YuriiKindrat (approved):

app/main.py
```diff
@@ -1 +1,43 @@
...
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        else:
+            return 0
```
MasterpieceElbow, line 16: You don't have to use `else` because you have a `return` statement in `if`. Just `return 0`.

MasterpieceElbow (changes requested):

app/main.py
BohdanKryven: Thanks! Fixed

app/main.py
BohdanKryven: Еhanks for the advice

katevasilenko (approved):

MasterpieceElbow (approved):

PR #74, opened 2022-05-26
Fix readme and tests

MasterpieceElbow:


tests/test_main.py
```diff
@@ -180,3 +195,14 @@ def test_dog_bring_slippers_method():
...
     assert main.feed_animals(animals) == total_food_points
+
+
+def test_feed_animals_should_feed_animal():
+    cat = main.Cat("Tom", is_hungry=True)
+    lion = main.Animal("Lion", appetite=25, is_hungry=True)
+    dog = main.Dog("Dog", is_hungry=True)
+    main.feed_animals([cat, lion, dog])
+
+    assert all(map(lambda x: x.is_hungry is False, [cat, lion, dog])), (
```
IvanRamyk, line 206: I prefer to use comprehensions instead of `map`:
```
assert all(not animal.is_hungry for animal in [cat, lion, dog])
```

IvanRamyk (approved):
Approved with suggestions

PR #83, opened 2022-06-18
Solution

muhactive:


app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        if self.is_hungry is True:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+        else:
+            self.appetite = 0
```
IvanRamyk, line 15: You shouldn't change the appetite of an animal when it's not hungry. This integer just shows a number of food points that are required to feed an animal IF it's hungry. If it isn't hungry no changes are required.

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #85, opened 2022-06-23
Solution

andrii-padalko:


app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name, is_hungry=True):
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers():
+        print("The slippers delivered!")
+
+
+def feed_animals(animals):
+    return sum([animal.feed() for animal in animals])
```
IvanRamyk, line 40:
```suggestion
    return sum(animal.feed() for animal in animals)
```
If you find the sum in this way Python wouldn't create a redundant list

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #105, opened 2022-08-02
Complete

Glasis9:


app/main.py
```diff
@@ -1 +1,41 @@
...
+    def __init__(self, name: str, appetite=7, is_hungry=True):
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers():
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list):
+    return sum([animal.feed() for animal in animals])
```
IvanRamyk, line 41:
```suggestion
    return sum(animal.feed() for animal in animals)
```
You could avoid creating an extra list

IvanRamyk (changes requested):

app/main.py
Glasis9: Fixed, Thanks for the tip

IvanRamyk (approved):
Well done!

PR #111, opened 2022-08-12
My Solution

YanPurdenko:


app/main.py
```diff
@@ -1 +1,39 @@
...
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        if self.is_hungry is True:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
DmytroSvirsa, line 17: You don’t need `else` if you have a `return` statement in `if`.

DmytroSvirsa (changes requested):

app/main.py
YanPurdenko: Changed

IvanRamyk (approved):
Well done!

PR #112, opened 2022-08-13
'Solution'

Pavilion4ik:


app/main.py
```diff
@@ -1 +1,38 @@
...
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
DmytroSvirsa, line 15: You don’t need `else` if you have a `return` statement in `if`.

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #113, opened 2022-08-14
'Solution'

Diliaver:


app/main.py
```diff
@@ -1 +1,55 @@
...
+        else:
+            self.is_hungry = False
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name, appetite=3, is_hungry=True):
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
```
DmytroSvirsa, line 26: You already use `super().__init__` from inherited Class `Animal` so it is code duplication:
```
self.name = name
self.appetite = appetite
self.is_hungry = is_hungry
```

app/main.py
```diff
@@ -1 +1,55 @@
...
+    @staticmethod
+    def catch_mouse():
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name, appetite=7, is_hungry=True):
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
```
DmytroSvirsa, line 40: You already use `super().__init__` from inherited Class `Animal` so it is code duplication:
```
self.name = name
self.appetite = appetite
self.is_hungry = is_hungry
```

app/main.py
```diff
@@ -1 +1,55 @@
...
+    def bring_slippers():
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals):
+    sum_feed_points = 0
+    for i in list_of_animals:
+        if i.is_hungry is True:
+            sum_feed_points += i.appetite
+        if i.is_hungry is True:
```
DmytroSvirsa, line 53: the same check, unite

DmytroSvirsa (changes requested):

app/main.py
IvanRamyk: You haven't corrected it

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #116, opened 2022-08-18
Solution

rurakite:


app/main.py
```diff
@@ -1 +1,43 @@
...
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True):
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
DmytroSvirsa, line 22: `Animal` class has `is_hungry` attribute, you can use it in `super()`

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    @staticmethod
+    def catch_mouse():
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True):
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
```
DmytroSvirsa, line 32: `Animal` class has `is_hungry` attribute, you can use it in `super()`

DmytroSvirsa (changes requested):

app/main.py
rurakite: Could you advise me how to do that please? I been trying and have no idea how to put it there((

PR #119, opened 2022-08-31
'Solution'

davidkrivko:


app/main.py
```diff
@@ -1 +1,44 @@
...
+    def bring_slippers():
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list):
+    amount = 0
+    for pet in animals:
+        if pet.is_hungry:
+            pet.feed()
+            amount += pet.appetite
```
IvanRamyk, line 42: use the value returned from the `feed` method to calculate the total amount. You wouldn't need to use `if` in that case

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        if self.appetite > 0 and self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
IvanRamyk, line 15: Please, remove the redundant else

IvanRamyk (changes requested):

PR #136, opened 2022-09-16
Solution

wQuelS:


Nattalli (approved):
Sooo good 😮

app/main.py
```diff
@@ -1 +1,39 @@
...
+    def __init__(self, name: str, is_hungry: bool = True):
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals])
```
wQuelS, line 39: actually, you don't need [] when using `sum` since this func can be applied on generator

PR #139, opened 2022-09-16
Solution

sbublinskyi:


app/main.py
```diff
@@ -1 +1,42 @@
-#  write your code here
+class Animal:
+
+    def __init__(self, name: str, appetite: int, is_hungry: bool = True):
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self):
```
wQuelS, line 8: Everything look really good, but add type annotations to all your functions.
If function returns nothing then it is -> None

wQuelS (changes requested):

wQuelS (approved):
Good job:)

app/main.py
```diff
@@ -1 +1,42 @@
...
+            return self.appetite
+
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name, is_hungry: bool = True):
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
TPolina, line 24: why don't you pass `is_hungry` to `super` as well?

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
```
TPolina, line 34: same here

TPolina (changes requested):

TPolina (approved):

PR #142, opened 2022-09-18
'Solution'

AVyha:


app/main.py
```diff
@@ -1 +1,45 @@
...
+            return 0
+        print(f"Eating {self.appetite} food points...")
+        self.is_hungry = False
+        points_eaten, self.appetite = self.appetite, 0
+        return points_eaten
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True):
+        super().__init__(name=name, appetite=0 if not is_hungry else 3)
```
mltdll, line 21: It would probably be better to pass `is_hungry` to `super().__init__()`, since you don't do anything to it.

mltdll (approved):

app/main.py
```diff
@@ -1 +1,45 @@
...
+class Animal:
+    def __init__(self, name: str, appetite: int, is_hungry: bool = True):
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
```
Y-Havryliv, line 10: add type annotations everywhere

app/main.py
```diff
@@ -1 +1,45 @@
...
+        if not self.is_hungry:
+            return 0
+        print(f"Eating {self.appetite} food points...")
+        self.is_hungry = False
+        points_eaten, self.appetite = self.appetite, 0
```
Y-Havryliv, line 15: you don't need to change `self.appetite` here

app/main.py
Y-Havryliv, line 21: `appetite` should equal 3, it's hungry by default

app/main.py
```diff
@@ -1 +1,45 @@
...
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def catch_mouse():
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True):
+        super().__init__(name=name, appetite=0 if not is_hungry else 7)
```
Y-Havryliv, line 32: `appetite` should equal 7, it's hungry by default

app/main.py
```diff
@@ -1 +1,45 @@
...
+        self.name = name
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers():
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list):
+    sum_of_points = 0
```
Y-Havryliv, line 42: it's up to you, but you can do it in a one line

app/main.py
Y-Havryliv: I agree

Y-Havryliv (changes requested):

Y-Havryliv (approved):

PR #147, opened 2022-09-20
Solution for the task

olia-trofymchuk:


app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
wQuelS, line 15: remove redundant else statement

app/main.py
```diff
@@ -1 +1,44 @@
...
+    def __init__(self, name: str, appetite: int, is_hungry: bool = True):
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
```
wQuelS, line 11: btw,  if self.is_hungry    is already boolean value, so u can get rid of "is True". Up to you

app/main.py
```diff
@@ -1 +1,44 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str, appetite: int, is_hungry: bool = True):
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self):
```
wQuelS, line 7: add type annotations to ALL your funcrionts. If function returns nothing then it is -> None

app/main.py
```diff
@@ -1 +1,44 @@
...
+    @staticmethod
+    def bring_slippers():
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    count = 0
+
+    for animal in animals:
+        animal.appetite = animal.feed()
```
wQuelS, line 41: you can make this in more pythonic way, by using list comprehension on gen.expression and sum() function. Up to you

wQuelS (changes requested):

app/main.py
olia-trofymchuk: Tests require to return 0 if animal isn't hungry

wQuelS: Yes, but u don't need esle to do so

olia-trofymchuk: Yes, exactly. Now I understand what you meant

wQuelS (approved):
Looks good to me

app/main.py
Y-Havryliv: yep, agree

Y-Havryliv (approved):
Well done!

PR #149, opened 2022-09-20
Solution

luxuriant777:


sbublinskyi (approved):
its good, but I recommend you to add annotations for methods, for example:
def print_name(self) -> None

Nattalli:
> its good, but I recommend you to add annotations for methods, for example: def print_name(self) -> None

You are right, type annotation is required. So you have to add them or your mentor will not approve these changes

luxuriant777:
type annotations were added

Nattalli (approved):
👍🏻

PR #150, opened 2022-09-21
Solution

taras-andruschenko:


sbublinskyi (approved):
👍

app/main.py
```diff
@@ -1 +1,49 @@
...
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
```
Nattalli, line 13: add type annotation for this function

app/main.py
```diff
@@ -1 +1,49 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    sum_of_food = 0
+    for animal in animals:
+        sum_of_food += animal.feed()
+    return sum_of_food
```
Nattalli, line 49: use function `sum`

Nattalli (changes requested):

app/main.py
taras-andruschenko: I thought it will be more "pythonic"

Nattalli: you can write it in one line - it will be more pythonic as well

app/main.py
Nattalli: and here: `-> int`

Nattalli (changes requested):

Nattalli (approved):
👍🏻

PR #153, opened 2022-09-21
Solution

yessloss:


app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self):
+        if self.is_hungry:
+            self.is_hungry = False
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        else:
```
IvanRamyk, line 15: Please, remove the redundant else

app/main.py
```diff
@@ -1 +1,49 @@
...
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name, is_hungry=True, appetite=3):
+        super().__init__(name, appetite, is_hungry)
+
+    def catch_mouse(self):
+        if self.is_hungry:
```
IvanRamyk, line 24: The method should print `The hunt began!` anyway not only if the cat is hungry

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name, is_hungry=True, appetite=7):
+        super().__init__(name, appetite, is_hungry)
+
+    def bring_slippers(self):
+        if self.is_hungry:
+            self.is_hungry = False
+            print("The slippers delivered!")
```
IvanRamyk, line 38: The same issue

app/main.py
```diff
@@ -1 +1,49 @@
...
+        else:
+            return 0
+
+
+def feed_animals(animals):
+    result = []
```
IvanRamyk, line 44: Use numeric variable to calculate `result`
```suggestion
    result = 0
```

app/main.py
```diff
@@ -1 +1,49 @@
...
+    for animal in animals:
+        if animal.is_hungry:
+            result.append(animal.appetite)
```
IvanRamyk, line 47: Use method `feed` here

IvanRamyk (changes requested):

app/main.py
yessloss: how I can use feed I in method?

IvanRamyk (changes requested):
You haven't provide any changes

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    def bring_slippers(self):
+        self.is_hungry = False
+        print("The slippers delivered!")
+
+
+def feed_animals(animals):
+    result = 0
+    for animal in animals:
+        if animal.is_hungry:
```
IvanRamyk, line 39: Use `feed` method to get appetite value for hungry animals

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1 +1,42 @@
...
+            result += animal.feed()
+            animal.is_hungry = False
```
IvanRamyk, line 41: You do it in the `feed` method in line 12. Setting `is_hungry` attribute to `False` one more time is extra

IvanRamyk (approved):
Well done!

PR #157, opened 2022-09-21
Solution)

Gr1zzZzzly:


app/main.py
```diff
@@ -1 +1,55 @@
...
+        appetite: int = 7
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(ls: list) -> int:
```
AVyha, line 54: in future try to use better name for argument like `animals` or `animal_list` not `ls`

AVyha (approved):

app/main.py
```diff
@@ -1 +1,55 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        else:
+            return(0)
```
Nattalli, line 22: just `return 0` will be looks better

app/main.py
Nattalli: and fix this, it's really good comment
don't use `ls`

Nattalli (changes requested):

Nattalli (approved):
Good job! 😸

PR #167, opened 2022-09-22
Solution

matechris:


app/main.py
```diff
@@ -1 +1,47 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=3, is_hungry=is_hungry)
+        self.name = name
```
HStakhiv, line 24: Do we need to store the arguments in the constructor?

app/main.py
```diff
@@ -1 +1,47 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    food_points = 0
+
+    for animal in animals:
+        food_points += animal.feed()
+
+    return food_points
```
HStakhiv, line 47: Good! But you can do it in one line (use `sum` and list comprehension).

HStakhiv (changes requested):
Good! But ... Why do you store the arguments in `Cat` and not in `Dog` as well?

HStakhiv (approved):
Nice!

Y-Havryliv (approved):
Nice!

PR #169, opened 2022-09-22
solve the task

Cerne13:


Kitazuka (approved):
Looks nice

app/main.py
```diff
@@ -1 +1,42 @@
...
+            name: str,
+            appetite: int,
+            is_hungry: bool = True
+    ):
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self):
+        print(f'Hello, I\'m {self.name}')
```
TPolina, line 13: use double quotes with f-string and don't use `\ '`

app/main.py
```diff
@@ -1 +1,42 @@
...
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f'Eating {self.appetite} food points...')
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name, is_hungry=True):
```
TPolina, line 24: add type annotation where it is missing

TPolina (changes requested):

app/main.py
Cerne13: Thank you!

TPolina (approved):
Well done!

PR #178, opened 2022-09-23
Solution

serviza:


app/main.py
```diff
@@ -1 +1,52 @@
...
+        else:
+            self.is_hungry = False
+            return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+        self.name = name
```
linaSam, line 27: this and next line of code are unnecessary, you can delete them

app/main.py
```diff
@@ -1 +1,52 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+        self.name = name
```
linaSam, line 39: the same issue - this line and 2 below can be deleted without influencing anything

linaSam (changes requested):

app/main.py
serviza: Agree with you! 
Thank you)

linaSam (approved):
Looks great to me!

app/main.py
```diff
@@ -1 +1,47 @@
...
+    def print_name(self):
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        else:
+            self.is_hungry = False
```
Nattalli, line 19: but it is already `self.is_hungry = False` so this line do nothing

app/main.py
```diff
@@ -1 +1,47 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food = 0
+    for animal in animals:
+        total_food += animal.feed()
+    return total_food
```
Nattalli, line 47: use `sum` function

Nattalli (changes requested):

app/main.py
serviza: Deleted

app/main.py
serviza: Done!

Nattalli (approved):
Good job!

PR #181, opened 2022-09-23
Created class Animal and count feed of inst

alexzubkoff:


app/main.py
```diff
@@ -1 +1,44 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry=True) -> None:
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
Mowka7, line 24: Why did you duplicated this? You can add "is_hungry" to super().__init__()

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry=True) -> None:
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
```
Mowka7, line 33: The same here. Actually I'm not sure would it be a mistake or not.. just keep in mind

Mowka7 (approved):

alexzubkoff:
I agree with you. I will remove that.

Сб, 24 сент. 2022 г. в 15:27, Mowka7 ***@***.***>:

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: Animal) -> int:
+    feed_sum = 0
+    for animal in animals_list:
+        if animal.is_hungry:
+            feed_sum += animal.feed()
```
Nattalli, line 41: use `sum` function

Nattalli (changes requested):

app/main.py
alexzubkoff: Thanks, for advice!

Nattalli (approved):
Good! ➕

PR #187, opened 2022-09-24
Feed animal first attempt

VitaliyBychevoy:


app/main.py
```diff
@@ -1 +1,60 @@
...
+        super().__init__(name, 7, is_hungry)
+        self.name = name
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Vasyl-Poremchuk, line 56: I think, you can use a one-liner like this `sum(animal.feed() for animal in animals)` in this method

app/main.py
```diff
@@ -1 +1,60 @@
...
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True and self.appetite != 0:
```
Vasyl-Poremchuk, line 18: I think this condition can be simplified, for example like this: `if self.is_hungry:`, because if this condition is True, it will be fulfilled

app/main.py
```diff
@@ -1 +1,60 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Vasyl-Poremchuk, line 22: And you don't need to use an `else`  condition, because if the previous condition is False, you just need to return 0. You can remove the `else`  condition and shift `return 0` 4 spaces to the left

Vasyl-Poremchuk (changes requested):

Vasyl-Poremchuk (approved):

app/main.py
```diff
@@ -1 +1,56 @@
 #  write your code here
```
Nattalli, line 1: Remove the unnecessary comment

app/main.py
```diff
@@ -1 +1,56 @@
...
+
+
+class Cat(Animal):
+
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
+            is_hungry:
+            bool = True
```
Nattalli, line 32: does it really needed to be on different lines?

Nattalli (changes requested):

PR #189, opened 2022-09-25
'Solution'

Menselface:


app/main.py
```diff
@@ -1 +1,42 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.appetite for animal in animals if animal.feed()])
```
FranchukS, line 42: You don`t need to use if condition in this comprehension.
Just use feed() method as output for each iteration

FranchukS (changes requested):

FranchukS (approved):
Nice!

Nattalli (changes requested):

Nattalli (approved):
Good job!

PR #206, opened 2022-10-09
'Solution'

IhorPokr:


app/main.py
```diff
@@ -1 +1,39 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals])
```
IvanRamyk, line 39:
```suggestion
    return sum(animal.feed() for animal in animals)
```
avoid creating an extra list

IvanRamyk (changes requested):

app/main.py
IhorPokr: > 

I fixed it)
I guess I accidentally hit "[" and didn't notice how it was added

IvanRamyk (approved):
Good job!

PR #209, opened 2022-10-10
Solution

GeorgePavlej:


app/main.py
```diff
@@ -1 +1,43 @@
...
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals])
```
IvanRamyk, line 43:
```suggestion
    return sum(animal.feed() for animal in animals)
```
avoid creating an extra list

IvanRamyk (changes requested):

app/main.py
GeorgePavlej: Ok. Thank you

IvanRamyk (approved):
Well done!

PR #217, opened 2022-10-22
Solution

ArtMot99:


app/main.py
```diff
@@ -1 +1,47 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
+            self.is_hungry = False
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        self.appetite = 0
+        return self.appetite
```
Abnormaltype, line 19:
```suggestion
    def feed(self) -> int:
        self.appetite = 0
        if self.is_hungry is True:
            self.is_hungry = False
            print(f"Eating {self.appetite} food points...")
            return self.appetite

        return self.appetite
```

Abnormaltype (approved):
would be better like that

app/main.py
ArtMot99: Hi, if I move the self.appetite variable, then I will fail the tests.

PR #222, opened 2022-10-29
Solution

Ida83:


app/main.py
```diff
@@ -1 +1,44 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list | Animal) -> int:
+    result = 0
+    for animal in animals:
+        food_points = animal.feed()
+        result += food_points
```
IvanRamyk, line 43:
```suggestion
        result += animal.feed()
```
you can avoid using an extra variable here

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #223, opened 2022-10-31
Solution

darynka-h:


app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> int:
+    sum_points = 0
+    for animal in animals_list:
+        if animal.feed():
```
IvanRamyk, line 44: use `feed` method to get number of the appetite points

IvanRamyk (changes requested):

app/main.py
darynka-h: What do you mean? In 44 string I use feed method to get number of the appetite each animal

darynka-h:
Changed solution

IvanRamyk (approved):
Well done!

PR #224, opened 2022-10-31
Solution

ochernyi:


app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    total_appetite = 0
+    for animal in list_of_animals:
+        if animal.is_hungry:
```
IvanRamyk, line 48: use `feed` method to get number of food points

IvanRamyk (changes requested):

app/main.py
ochernyi: what exactly do you mean? This method return total number of food points for all animals in the list. It's all what I need

ochernyi: > use `feed` method to get number of food points
what exactly do you mean? This method return total number of food points for all animals in the list. It's all what I need

IvanRamyk: instead of checking the `is_hungry` attribute directly and  getting the `appetite`  attribute value just use `feed` method

IvanRamyk (changes requested):

DmytroSvirsa (approved):

PR #230, opened 2022-11-10
Solution

yuraua5:


app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> int:
+    return sum([animal.appetite and animal.feed()
+                for animal in animals_list
+                if animal.is_hungry is True])
```
DmytroSvirsa, line 47: You can leave just animal.feed() without cheking  if animal.is_hungry is True - you already do it in animal.feed()

DmytroSvirsa (changes requested):

app/main.py
```diff
@@ -1 +1,46 @@
...
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, appetite)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> int:
+    return sum([animal.appetite and animal.feed()
```
DmytroSvirsa, line 45: why you put this statement - `animal.appetite and animal.feed()     `
and you can remove `[]` - `return sum(animal.feed() for animal in animals_list)`

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #234, opened 2022-11-15
Solution

MykytaMoshchenko:


app/main.py
```diff
@@ -1 +1,45 @@
...
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> int:
+    result = 0
+    for animal in animals_list:
+        if animal.is_hungry is True:
+            result += animal.appetite
+            animal.feed()
```
Abnormaltype, line 44:
```suggestion
        result += animal.feed()
            
```

Abnormaltype (changes requested):

app/main.py
MykytaMoshchenko: Thank you for the better option! 
The solution is updated.

Abnormaltype (approved):
Well done!

PR #237, opened 2022-11-18
'Solution'

Sebshe:


app/main.py
```diff
@@ -1 +1,57 @@
...
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
```
OleksandrPozden, line 16: redundant check. it is better to write ```if self.is_hungry:``` instead

OleksandrPozden (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #239, opened 2022-11-20
Feed animals solution

Polyakiv-Andrey:


app/main.py
```diff
@@ -1 +1,51 @@
...
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
```
OleksandrPozden, line 13: redundant method. it never used

app/main.py
```diff
@@ -1 +1,51 @@
...
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
```
OleksandrPozden, line 17: the redundant `is True`

app/main.py
```diff
@@ -1 +1,51 @@
...
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
OleksandrPozden, line 29: is_hungry should be passed to the parent's constructor

app/main.py
```diff
@@ -1 +1,51 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
```
OleksandrPozden, line 40: here as well

OleksandrPozden (changes requested):

app/main.py
Polyakiv-Andrey: Without this method 12 errors

Polyakiv-Andrey: This method is required by the task

Abnormaltype (approved):
Well done!

PR #247, opened 2022-11-28
Develop

T-Vlad-96:


app/main.py
```diff
@@ -1 +1,42 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str, appetite: int, is_hungry: bool = True) \
+            -> None:
```
YuriiKindrat, line 3: fix breaking line here according to the best practices

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,43 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str,
```
OleksandrPozden, line 2: add a break line here

OleksandrPozden (changes requested):

PR #257, opened 2022-12-07
'Solution'

evgenijmartynuk07:
cat = Cat("Cat", False)
cat.appetite == False and cat.is_hungry == True
Animal.feed() = output((f"Eating False food points...")) and tests works correctly.
I could not understand for myself what I was doing wrong. but then I looked at the tests

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -12,12 +12,11 @@ class Animal:
         print(f"Hello, I'm {self.name}")
 
     def feed(self) -> int:
-        if self.is_hungry is False:
+        if self.is_hungry is False or self.appetite is False:
             return 0
-        else:
-            print(f"Eating {self.appetite} food points...")
-            self.is_hungry = False
-            return self.appetite
+        print(f"Eating {self.appetite} food points...")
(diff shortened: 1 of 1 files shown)


evgenijmartynuk07:
> cat = Cat("Cat", False) cat.appetite == False and cat.is_hungry == True cat.feed() = output((f"Eating False food points...")) and tests works correctly. I could not understand for myself what I was doing wrong. but then I looked at the tests

app/main.py
```diff
@@ -1 +1,46 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is False:
+            return 0
+        else:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
```
Abnormaltype, line 20: write without else

Abnormaltype (changes requested):

Abnormaltype (approved):

PR #263, opened 2022-12-12
Solution

Isles22:


app/main.py
```diff
@@ -1 +1,56 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+
+        if self.is_hungry is True:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Serhii-Leonenko, line 21: Remove redundant else here.

app/main.py
```diff
@@ -1 +1,56 @@
...
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+        self, name: int,
+        is_hungry: bool = True,
+        appetite: int = 7
+    ) -> None:
```
Serhii-Leonenko, line 44:
```suggestion
    def __init__(
        self, 
        name: int,
        is_hungry: bool = True,
        appetite: int = 7
    ) -> None:
```

Fix line wrapping, also in other similar cases

Serhii-Leonenko (changes requested):

app/main.py
Isles22: got it thanks

Serhii-Leonenko (approved):
Good job!

PR #279, opened 2022-12-31
Solution

serhii-skivka:


app/main.py
```diff
@@ -1 +1,48 @@
-#  write your code here
+class Animal:
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True
+                 ) -> None:
```
OleksandrPozden, line 6:
```suggestion
    def __init__(
             self,
                 name: str,
                 appetite: int,
                 is_hungry: bool = True
        ) -> None:
```
re-format it accordingly to the best practices PEP 08

app/main.py
```diff
@@ -1 +1,48 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True) -> None:
```
OleksandrPozden, line 25: here as well

app/main.py
```diff
@@ -1 +1,48 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
```
OleksandrPozden, line 37: and here

OleksandrPozden (changes requested):

YuriiKindrat (approved):

PR #280, opened 2023-01-01
'Solution'

artur-leleiko:


app/main.py
```diff
@@ -1 +1,42 @@
-#  write your code here
+class Animal:
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True
+                 ) -> None:
```
OleksandrPozden, line 6:
```suggestion
    def __init__(
            self,
                 name: str,
                 appetite: int,
                 is_hungry: bool = True
    ) -> None:
```
reformat it according to the best practices pep 8

app/main.py
```diff
@@ -1 +1,42 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
OleksandrPozden, line 41: add more explicit type annotation

OleksandrPozden (changes requested):

YuriiKindrat (approved):

PR #300, opened 2023-01-17
Solution

avkpol:


app/main.py
```diff
@@ -1 +1,52 @@
...
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            if self.appetite != 0:
```
IvanRamyk, line 14: Use `and` instead of two nested ifs. it makes code more readable
DO NOT:
```
if condition1:
    if condition2:
         code ... 
```
DO:
```
if condition1 and consition2:
    code ....
```

app/main.py
```diff
@@ -1 +1,52 @@
...
+        super().__init__(name, appetite, is_hungry)
+        self.appetite = appetite
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    points = []
```
IvanRamyk, line 48: you don't really need to create a list to calculate a sum. try to do it using just a number

IvanRamyk (changes requested):

app/main.py
avkpol: I don't understand how to do that in another way, it doesn' t work like this:
 # points = 0
    # for animal in animals:
    #     points += animal.feed()
    #     return points
I refactored the code as a one-liner: return sum(animal.feed() for animal in animals)

ArsenPidhoretskyi (approved):
Well done

PR #307, opened 2023-01-24
feed animals

megah1t:


app/main.py
```diff
@@ -1 +1,40 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
ArsenPidhoretskyi, line 24: remove this method
it will automatically call parent

ArsenPidhoretskyi: And result will be the same

ArsenPidhoretskyi: The same with Dog

ArsenPidhoretskyi (changes requested):

app/main.py
megah1t: Если я правильно понял

app/main.py
```diff
@@ -1 +1,42 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str,
+                 appetite: int,
+                 is_hungry: bool = True
+                 ) -> None:
```
katevasilenko, line 5:
```suggestion
    def __init__(
          self, 
          name: str,
          appetite: int,
          is_hungry: bool = True
    ) -> None:
```
Please be patient with indentation. This is the right way to split line

app/main.py
```diff
@@ -1 +1,42 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = 3
+        self.is_hungry = is_hungry
```
katevasilenko, line 25: Actually you need super here, and you need to inherit is_hungry as well.

app/main.py
```diff
@@ -1 +1,42 @@
...
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
```
katevasilenko, line 22: Set appetite with default value here

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = 7
+        self.is_hungry = is_hungry
```
katevasilenko, line 35: Do the same way as in Cat class. Set parameter appetite with default value and use super with all attributes

katevasilenko (changes requested):

ArsenPidhoretskyi (approved):
Well done

PR #308, opened 2023-01-25
'Solution'

andy77andy:


Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,53 @@
...
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([Animal.feed(animal) for animal in animals])
```
olena-hul, line 53: `[]` is not needed here, just `sum(Animal.feed(animal) for animal in animals)` will be enough

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,53 @@
...
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum(Animal.feed(animal) for animal in animals)
```
ArsenPidhoretskyi, line 53: Animal.feed(animal) is wrong construction
We can call to instance this method because it is a simple method, not a classmethod

ArsenPidhoretskyi (changes requested):

app/main.py
andy77andy: done)

ArsenPidhoretskyi (approved):
Well done!

PR #310, opened 2023-01-26
Solution

Andrew13X:


ArsenPidhoretskyi (approved):
Well done.
Also, feed_animals can be re-writen by usage of sum function

PR #318, opened 2023-02-01
Solution

oleksandr-lysanets:


app/main.py
```diff
@@ -1 +1,46 @@
...
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        else:
```
ArsenPidhoretskyi, line 19: else: is redundant here

app/main.py
```diff
@@ -1 +1,46 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: [Animal]) -> int:
+    food = 0
+    for animal in animals:
+        food += animal.feed()
+        animal.is_hungry = False
+    return food
```
ArsenPidhoretskyi, line 46: use sum() function

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,42 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: [Animal]) -> int:
```
ArsenPidhoretskyi, line 41:
```suggestion
def feed_animals(animals: list[Animal]) -> int:
```
If not you can use:
from typing import List and than List[Animal]

ArsenPidhoretskyi (approved):
Good job!

PR #319, opened 2023-02-02
Solution

Rename925:


app/main.py
```diff
@@ -1 +1,63 @@
...
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, is_hungry)
+        self.name = name
+        self.is_hungry = is_hungry
+        self.appetite = 3
```
katevasilenko, line 35: You don’t those after super(). Also set appetite with default value in super too

app/main.py
```diff
@@ -1 +1,63 @@
...
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, is_hungry)
+        self.name = name
+        self.is_hungry = is_hungry
+        self.appetite = 7
```
katevasilenko, line 51: The same as in Cat class

app/main.py
```diff
@@ -1 +1,63 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
katevasilenko, line 58: Please write this function using `sum()`

katevasilenko (changes requested):

app/main.py
ArsenPidhoretskyi: Really you do not need to check for animal.is_hungry

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Well Done 👍

PR #322, opened 2023-02-05
'Solution'

natalie-goriela:


app/main.py
```diff
@@ -1 +1,56 @@
...
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            food_points = self.appetite
+            self.is_hungry = False
+            self.appetite = 0
+            return food_points
+        else:
```
ArsenPidhoretskyi, line 25: else is redundant here

app/main.py
```diff
@@ -1 +1,56 @@
...
+            self.appetite = 0
```
ArsenPidhoretskyi, line 26: > feed - should print Eating {appetite} food points..., set is_hungry to False and return number of eaten food points if animal is hungry. Otherwise, it should return 0 and print nothing.

No info to set appetite to 0, just return 0

app/main.py
```diff
@@ -1 +1,56 @@
...
+            return self.appetite
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+        self.name = name
+        self.is_hungry = is_hungry
```
ArsenPidhoretskyi, line 34: No need to do like this. You can pass all parameters with to `super().__init__()`

ArsenPidhoretskyi: The same in dog class

ArsenPidhoretskyi: https://realpython.com/python-super/#super-in-single-inheritance

app/main.py
```diff
@@ -1 +1,56 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+
+    sum_of_food = sum([animal.feed() for animal in animals])
```
ArsenPidhoretskyi, line 54: No need to convert to list.
```suggestion
    sum_of_food = sum(animal.feed() for animal in animals)
```

ArsenPidhoretskyi (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #324, opened 2023-02-07
Solution

MaybeTI:


app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> int:
+    return sum([animal.feed() for animal in animals_list if animal.is_hungry])
```
manjustice, line 40: You don't need 'if' here

manjustice (changes requested):

app/main.py
MaybeTI: Thank you, I forget it

manjustice (approved):
Well done!

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> int:
+    return sum([animal.feed() for animal in animals_list])
```
wQuelS, line 40: you don't need [], 'cause `sum` can work with generators

wQuelS (changes requested):

NazarHladaniuk (approved):

YuriiKindrat (approved):

PR #326, opened 2023-02-07
Develop

AlimFreight:


app/main.py
```diff
@@ -1 +1,55 @@
...
+
+
+class Cat(Animal):
+
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True
+                 ) -> None:
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
ArsenPidhoretskyi, line 33: You must pass this parameter into super

app/main.py
```diff
@@ -1 +1,55 @@
...
+
+
+class Dog(Animal):
+
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True
+                 ) -> None:
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
```
ArsenPidhoretskyi, line 47: You must pass this parameter into super

ArsenPidhoretskyi (changes requested):

app/main.py
AlimFreight: What do you mean? Which parameter? If you mean "is_hungry" then it doesn't work this way. Could you point more precisely please?

vasylhnatiuk (approved):
Good job!

PR #327, opened 2023-02-08
Solution

Roman28101:


app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
ArsenPidhoretskyi, line 21: else: redundant here

app/main.py
```diff
@@ -1 +1,49 @@
...
+            return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
ArsenPidhoretskyi, line 29: pass this parameter into the super class

ArsenPidhoretskyi: The same with dog

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    food = 0
+    for animal in animals:
+        food += animal.feed()
+    return food
```
ArsenPidhoretskyi, line 49: Use sum() function

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Well Done 👍

PR #328, opened 2023-02-08
Solution

SafonovVladimir:
add my solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,48 @@
-#  write your code here
+class Animal:
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
+
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            self.is_hungry = False
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Dog, self).__init__(name, appetite=7)
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> int:
+    result = 0
+    for animal in animals_list:
+        if animal.is_hungry:
+            animal.feed()
+            result += animal.appetite
+    return result


app/main.py
```diff
@@ -1 +1,48 @@
...
+            self.is_hungry = False
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
AndereLion, line 25: You can write "is_hungry" right away to the super constructor

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Dog, self).__init__(name, appetite=7)
+        self.is_hungry = is_hungry
```
AndereLion, line 35: Same =)

AndereLion (changes requested):

app/main.py
SafonovVladimir: Done!

AndereLion (approved):

app/main.py
```diff
@@ -1 +1,43 @@
...
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Dog, self).__init__(name, appetite=7, is_hungry=is_hungry)
```
wQuelS, line 32: Now, this function can be used `super().__init__(..)` and it will automatically pass `Dog, self`

app/main.py
```diff
@@ -1 +1,43 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> int:
+    result = 0
+    for animal in animals_list:
+        result += animal.feed()
+    return result
```
wQuelS, line 43: You can use `sum`. Make it in one line

wQuelS (changes requested):

app/main.py
SafonovVladimir: Of couse, using list comprehension.
Done

app/main.py
wQuelS: You still pass `Dog, self` to `super()` explicitly. In last updates `super()` obtained possibility to automatically take arguments of the class in which this function is used

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Dog, self).__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> int:
+    return sum([animal.feed() for animal in animals_list])
```
wQuelS, line 40: `sum` can work with generators, so [] are not needed. Generator is faster than comprehension, so it would be better to use it here

wQuelS (changes requested):

app/main.py
SafonovVladimir: Done

wQuelS (approved):
Very good

PR #329, opened 2023-02-08
1 commit

AndereLion:


app/main.py
```diff
@@ -1 +1,42 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str, appetite: int,
+                 is_hungry: bool = True) -> None:
```
SafonovVladimir, line 3: Move appetite to new line

SafonovVladimir (changes requested):

SafonovVladimir (approved):
GJ

app/main.py
```diff
@@ -1 +1,43 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    sum_of_feeding = 0
+    for animal in list_of_animals:
+        sum_of_feeding += animal.feed()
+    return sum_of_feeding
```
wQuelS, line 43: use built-in function called `sum`
Do it in one line

wQuelS (changes requested):

app/main.py
AndereLion: Fixed.

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    return sum([animal.feed() for animal in list_of_animals])
```
wQuelS, line 40: `sum` can work with generators, so [] are not needed. Generator is faster than comprehension, so it would be better to use it here

wQuelS (changes requested):

app/main.py
AndereLion: Done

app/main.py
```diff
@@ -1 +1,40 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str,
```
YuriiKindrat, line 2: fix indentation here, one arg per line

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    return sum((animal.feed() for animal in list_of_animals))
```
YuriiKindrat, line 40: no need to use double brackets here

YuriiKindrat (changes requested):

app/main.py
AndereLion: Fixed

app/main.py
AndereLion: Fixed

YuriiKindrat (approved):

PR #331, opened 2023-02-08
Solution

DavidGlusch:


app/main.py
```diff
@@ -1 +1,45 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=True)
+        self.name = name
+        self.is_hungry = is_hungry
```
NEkropo1, line 26:
```suggestion
```
Since You already inherited attributes from Animal, these lines are redundant.

app/main.py
```diff
@@ -1 +1,45 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=True)
+        self.name = name
+        self.is_hungry = is_hungry
```
NEkropo1, line 37:
```suggestion
```
Same here

NEkropo1 (changes requested):

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
NEkropo1, line 45: It's just simplification with not much impact, but, since you know it, this code works 1 bit faster )
```suggestion
    return sum(animal.feed() for animal in animals)
```

NEkropo1 (approved):
Nice, simple and efficient!

app/main.py
```diff
@@ -1 +1,41 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
wQuelS, line 41: `sum` can work with generators, so [] are not needed

wQuelS (changes requested):

PR #334, opened 2023-02-09
solution

oksana-feshchenko:


app/main.py
```diff
@@ -1 +1,41 @@
...
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            eaten = self.appetite
+            self.is_hungry = False
+        return eaten
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, is_hungry=is_hungry, appetite=3)
```
kuznetsb, line 23: if u use named parameters I think it's better to pass name=name also, but I'm not confident)

kuznetsb (approved):
Well done)

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def print_name(self) -> str:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        eaten = 0
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            eaten = self.appetite
+            self.is_hungry = False
```
wQuelS, line 17: `self.is_hungry = not self.is_hungry` is more reliable

app/main.py
wQuelS, line 23: pass `name=name` also, if u use named params

app/main.py
```diff
@@ -1 +1,41 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, is_hungry=is_hungry, appetite=3)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, is_hungry=is_hungry, appetite=7)
```
wQuelS, line 31: same

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    food = 0
+    for animal in animals:
+        food += animal.feed()
+    return food
```
wQuelS, line 41: It can be done in one line with `sum` function

wQuelS (changes requested):

YuriiKindrat (approved):

PR #335, opened 2023-02-09
Solution

kuznetsb:


app/main.py
```diff
@@ -1 +1,40 @@
...
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
```
oksanaaam, line 23: great solution!

oksanaaam (approved):

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
wQuelS, line 40: actually, you don't need [] here, 'cause `sum` can work with generators

wQuelS (changes requested):

PR #337, opened 2023-02-09
Solution

ZabFTFT:


app/main.py
```diff
@@ -1 +1,60 @@
...
+        Animal.list_of_animals.append(self)
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = not self.is_hungry
+            food_points_holder = self.appetite
```
Maytreyaya, line 23: we don't need "food points holder", we can just return appetite

app/main.py
```diff
@@ -1 +1,60 @@
...
+            self.appetite = 0
```
Maytreyaya, line 24: we don't have to chage appetite of animal, we should change only if its hungry or not

app/main.py
```diff
@@ -1 +1,60 @@
...
+
+class Cat(Animal):
+
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 3
+    ) -> None:
+        super().__init__(name=name, appetite=appetite, is_hungry=is_hungry)
```
Maytreyaya, line 37: its no use to write like this u can write just `(name, appetite, is_hungry)`

app/main.py
```diff
@@ -1 +1,60 @@
...
+
+class Dog(Animal):
+
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 7
+    ) -> None:
+        super().__init__(name=name, appetite=appetite, is_hungry=is_hungry)
```
Maytreyaya, line 52: same here

Maytreyaya (changes requested):

app/main.py
ZabFTFT: Thank you very much for review!

app/main.py
```diff
@@ -1 +1,58 @@
...
+
+class Cat(Animal):
+
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 3
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
```
yura-hudzovskyi, line 35: **`Note: you need call the super class constructor with appetite equal to 3.`** you should set default value in super, not in constructor

app/main.py
```diff
@@ -1 +1,58 @@
...
+
+class Dog(Animal):
+
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 7
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
```
yura-hudzovskyi, line 50:
```
The last class you should implement is a Dog class. Its constructor should have two arguments:

name - the name of a dog
is_hungry - with True default value
All dogs should have appetite equal to 7.
```
same

yura-hudzovskyi (changes requested):

app/main.py
Maytreyaya: you are not right. i don't think you can put default value in super. check information about it one more time, please. and  by the way i've already cheked it, so next to check is mentor

app/main.py
ZabFTFT: what's the difference? All dogs have same appetite value. We have same result with different approach.

yura-hudzovskyi (approved):

app/main.py
```diff
@@ -1 +1,56 @@
...
+    ) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in list_of_animals])
```
wQuelS, line 56: you can use `sum` without []

wQuelS (changes requested):

wQuelS (approved):
Nice

PR #338, opened 2023-02-09
Animal Solution

Maytreyaya:


app/main.py
```diff
@@ -1 +1,58 @@
...
+                 appetite: int,
+                 is_hungry: bool = True) -> Animal:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> str:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> str:
```
Zenionoi, line 16: `return number of eaten food points`, change type annotation to `int`

app/main.py
```diff
@@ -1 +1,58 @@
...
+    def bring_slippers() -> str:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    result_list = []
+    for animal in list_of_animals:
+        if animal.is_hungry:
+            animal.feed()
+            result_list.append(animal.appetite)
```
Zenionoi, line 57: Function `animal.feed()` will return animal's appetite if `animal.is_hungry`, so you don`t need to duplicate and can result of func at once

Zenionoi (changes requested):

Zenionoi (approved):

app/main.py
```diff
@@ -1 +1,57 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True,
+                 appetite: int = 3) -> Cat:
```
wQuelS, line 28: you don't need it to pass in `init`. You have `super` func for such things.

app/main.py
```diff
@@ -1 +1,57 @@
...
+        super().__init__(name, appetite, is_hungry)
+        self.name = name
+        self.is_hungry = is_hungry
```
wQuelS, line 31: those are redundant too

app/main.py
```diff
@@ -1 +1,57 @@
...
+
+
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True,
+                 appetite: int = 7) -> Dog:
+        super().__init__(name, appetite, is_hungry)
+        self.name = name
+        self.is_hungry = is_hungry
```
wQuelS, line 45: same here

app/main.py
```diff
@@ -1 +1,57 @@
...
+    def bring_slippers() -> str:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    result_list = []
+    for animal in list_of_animals:
+        if animal.is_hungry:
+            result_list.append(animal.feed())
+    return sum(result_list)
```
wQuelS, line 57: can you use `sum` in such a way that this code becomes a oneliner?

wQuelS (changes requested):

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    @staticmethod
+    def bring_slippers() -> str:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    return sum([animal.feed()
+                for animal in list_of_animals
+                if animal.is_hungry])
```
YuriiKindrat, line 51: no need to use [] brackets here, sum can work with generators

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    @staticmethod
+    def bring_slippers() -> str:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    return sum(animal.feed()
+               for animal in list_of_animals
+               if animal.is_hungry)
```
wQuelS, line 51: this check is redundant, since `feed` method has it's own similar check

wQuelS (changes requested):

wQuelS (approved):
Good

PR #339, opened 2023-02-09
Solution

yura-hudzovskyi:


DmytryTong (approved):
I do not know about the @staticmethod, but it this particular case, it's should be fine) And in a super() you may using just "is_hungry" not a "is_hungry=is_hungry" cause the value is already were in a parent __init__. But is still worfking, so - I don't mind) Nice jod)

app/main.py
```diff
@@ -1 +1,45 @@
...
+            print(f"Eating {self.appetite} food points...")
+            food_points = self.appetite
+            self.is_hungry = not self.is_hungry
+            return food_points
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Cat, self).__init__(name, appetite=3, is_hungry=is_hungry)
```
wQuelS, line 25: You don't really need to pass `Cat, self` in `super` explicitly. With last updates, it can be used like this `super` and it will automatically take those params

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Dog, self).__init__(name, appetite=7, is_hungry=is_hungry)
```
wQuelS, line 34: same

app/main.py
```diff
@@ -1 +1,45 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    count_of_feed_point = 0
+    for animal in animals:
+        count_of_feed_point += animal.feed()
+    return count_of_feed_point
```
wQuelS, line 45: Can you try to use `sum` and make it in one line.
Small hint: `sum` can work with generators

wQuelS (changes requested):

wQuelS (approved):
Great

PR #340, opened 2023-02-09
Solution

Zenionoi:


amber-marichi (approved):
looks pretty. approved

app/main.py
```diff
@@ -1 +1,50 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    count = 0
+    for animal in animals:
+        count += animal.feed()
+
+    return count
```
wQuelS, line 50: Can you make it in one line using `sum` ?

wQuelS (changes requested):

app/main.py
```diff
@@ -1 +1,46 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
wQuelS, line 46: actually, [] are not needed here, since `sum` can work with generators

wQuelS (changes requested):

wQuelS (approved):
Nice

PR #341, opened 2023-02-09
Solution DonchykD

DmytryTong:


app/main.py
```diff
@@ -1 +1,47 @@
...
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
```
SergiiMachulin, line 15: You can write here condition without ```is True```, because it is set by default

app/main.py
```diff
@@ -1 +1,47 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
```
SergiiMachulin, line 23: Why do you use here new line?)

app/main.py
```diff
@@ -1 +1,47 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    @classmethod
```
SergiiMachulin, line 27: according to logic and to checklist in README.md it is better to use ```@staticmethod``` here and below

app/main.py
```diff
@@ -1 +1,47 @@
...
+    def catch_mouse(cls) -> str:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @classmethod
+    def bring_slippers(cls) -> str:
```
SergiiMachulin, line 37: print() return None not string

app/main.py
```diff
@@ -1 +1,47 @@
...
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    @classmethod
+    def catch_mouse(cls) -> str:
```
SergiiMachulin, line 28: print() return None not str

app/main.py
```diff
@@ -1 +1,47 @@
...
+        super().__init__(name, 7, is_hungry)
+
+    @classmethod
+    def bring_slippers(cls) -> str:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    summ = 0
+    for animal in list_of_animals:
```
SergiiMachulin, line 43: according to checklist in README.md it's better to use list comprehension here

app/main.py
```diff
@@ -1 +1,47 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @classmethod
+    def bring_slippers(cls) -> str:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
```
SergiiMachulin, line 41: You could add more detailed annotation of class inside list

SergiiMachulin (changes requested):

app/main.py
DmytryTong: Fixed.

app/main.py
DmytryTong: It seemed to me, that new line there makes code easier to read, but it's no necessary.

app/main.py
DmytryTong: Ok, got it - fixed.

app/main.py
DmytryTong: Fixed.

app/main.py
DmytryTong: Fixed.

app/main.py
DmytryTong: It was rather recomendation, but I get it - fixed.

app/main.py
DmytryTong: Done.

app/main.py
```diff
@@ -12,36 +12,35 @@ def print_name(self) -> None:
         print(f"Hello, I'm {self.name}")
 
     def feed(self) -> int:
-        if self.is_hungry is True:
-            print(f"Eating {self.appetite} food points...")
-            self.is_hungry = False
-            return self.appetite
-        return 0
+        if self.is_hungry is False:
```
SergiiMachulin, line 15: You previous code was ok, I only suggest to do condition simplier like ```if self.is_hungry:```

app/main.py
```diff
@@ -12,36 +12,35 @@ def print_name(self) -> None:
...
 
 
-def feed_animals(list_of_animals: list) -> int:
+def feed_animals(list_of_animals: list[Animal]) -> int:
     summ = 0
-    for animal in list_of_animals:
-        if animal.is_hungry is True:
-            animal.feed()
-            summ += animal.appetite
+    summ = sum([animal.appetite
```
SergiiMachulin, line 42: For what reason do you duplicate your code, and also create variable for it? You can put all in one line ```return sum([])```

SergiiMachulin (changes requested):

app/main.py
DmytryTong: Ok, I guess got it wrong - fixed.

app/main.py
DmytryTong: Ok, got it - fixed.

SergiiMachulin (approved):
Nice!

app/main.py
```diff
@@ -1 +1,44 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list[Animal]) -> int:
+
+    return sum([animal.feed()
+                for animal in list_of_animals
+                if animal.is_hungry is True])
```
wQuelS, line 44: first: `sum` can work with generators, so [] are not needed
second: `if animal.is_hungry is True` is redundant, 'cause `feed()` method has the same check

wQuelS (changes requested):

app/main.py
DmytryTong: Fixed.

wQuelS (approved):
Good job

PR #342, opened 2023-02-09
solution

amber-marichi:


app/main.py
```diff
@@ -1 +1,50 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> None:
+        food_points = 0
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            food_points += self.appetite
+        return food_points
```
Danil228, line 25: "set is_hungry to False and return number of eaten food points if animal is hungry. Otherwise, it should return 0 and print nothing"
You dont need food_points variable, you just need to return self.appetite if animal is hungry, else return 0

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list[Animal]) -> int:
+    total_points = 0
+    for animal in animal_list:
+        total_points += animal.feed()
+    return total_points
```
Danil228, line 50: you can do this in one line, but it a matter of taste

Danil228 (changes requested):

Danil228 (approved):

wQuelS (approved):
Great

PR #343, opened 2023-02-10
Solution

SergiiMachulin:


MikhailLyvak (approved):
Well done!

app/main.py
```diff
@@ -1 +1,47 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
wQuelS, line 29: just pass `is_hungry into `super()`

app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
```
wQuelS, line 39: same

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
```
wQuelS, line 21: I guess `self.is_hungry = not self.is_hungry` is more reliable

app/main.py
```diff
@@ -1 +1,47 @@
...
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals if animal.is_hungry])
```
wQuelS, line 47: first: you don't need [], since `sum` can work with generators
second: `if animal.is_hungry` is redundant, since `feed` has it's own check

wQuelS (changes requested):

app/main.py
SergiiMachulin: According to README.md Class Cat have only two arguments(name, is_hungry), and we need call the super class constructor with appetite = ... pytest doesnt work if opposite

app/main.py
SergiiMachulin: But here it is possible if we change places of appetite argument and is_hungry

app/main.py
wQuelS: Just try to pass `is_hungry` in `super()` func :))

wQuelS (changes requested):

me1nyk (approved):
Well done

YuriiKindrat (approved):

PR #344, opened 2023-02-10
Solution

Danil228:


app/main.py
```diff
@@ -1 +1,43 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
Morty67, line 21: you can skip else and use return

Morty67 (approved):

Morty67 (approved):

app/main.py
```diff
@@ -1 +1,42 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: List[Animal]) -> int:
+    return sum([animal.feed() for animal in animals_list])
```
wQuelS, line 42: `sum` can work with generators, so [] are not needed

wQuelS (changes requested):

app/main.py
Danil228: Done

wQuelS (approved):
Great

PR #345, opened 2023-02-10
Solution

MikhailLyvak:


inrock1 (approved):
Nice

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> None:
+        if self.is_hungry is True:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
wQuelS, line 18: `else` is redundant here

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> None:
+    return sum([animal.feed() for animal in animals_list if animal.is_hungry])
```
wQuelS, line 40: `sum` can work with generators, so there is no need in []

this check `if animal.is_hungry` is redundant, 'cause `feed` method has already it

wQuelS (changes requested):

app/main.py
MikhailLyvak: good advice. Thanks)

app/main.py
MikhailLyvak: Wouldn't that be a bad tone?

YuriiKindrat (approved):

PR #346, opened 2023-02-10
Solution

WalDragnir:


app/main.py
```diff
@@ -1 +1,47 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=3, is_hungry=is_hungry)
```
Grayder0152, line 25: More correct create constant for default cat appetite and use it instead 3

app/main.py
```diff
@@ -1 +1,47 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
```
Grayder0152, line 36: Same as above

Grayder0152 (approved):

app/main.py
WalDragnir: image
With constant dosnt work

PR #347, opened 2023-02-10
'Solution'

Morty67:


app/main.py
```diff
@@ -1 +1,39 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals])
```
AlessaQ, line 39: Square brackets are not necessary

AlessaQ (approved):

app/main.py
wQuelS: True

wQuelS, line 39: `sum` can work with generators which a faster than comprehensions

wQuelS (changes requested):

wQuelS (approved):
Nice work

PR #348, opened 2023-02-10
'Solution1'

inrock1:


app/main.py
```diff
@@ -1 +1,49 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(list_animals: list) -> int:
+    food_points = 0
+    for animal in list_animals:
+        if animal.is_hungry:
+            animal.feed()
+            food_points += animal.appetite
+    return food_points
```
viktoria-rybenchuk, line 49: Don't use a for loop if you can use a list comprehension.

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,49 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
viktoria-rybenchuk, line 18: Don't use an extra else statement.

app/main.py
inrock1: done

app/main.py
inrock1: done

viktoria-rybenchuk (approved):

YuriiKindrat (approved):

PR #349, opened 2023-02-10
Solution

ZhAlexR:


oigod (approved):

app/main.py
```diff
@@ -1 +1,57 @@
...
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> None | int:
```
usrofgh, line 17: I'm not sure about None here. Could you please explain why do you use it cause i can't understand

ZhAlexR: You are right! 
I just forgot to delete it. 
The previously was supposed to  return None. So "None" is just legacy.
I will fix

app/main.py
```diff
@@ -1 +1,57 @@
...
+                         appetite=appetite,
+                         is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
Anatolii-Poznyak, line 57:
```suggestion
    return sum(animal.feed() for animal in animals)
```
you could use sum w/o [], like in case sum(range(5))

app/main.py
```diff
@@ -1 +1,57 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
Anatolii-Poznyak, line 23: cmooon, maaaan )))) 
Screenshot 2023-02-10 at 18 49 46

Anatolii-Poznyak (approved):

app/main.py
wQuelS: agree

app/main.py
wQuelS: agree

app/main.py
```diff
@@ -1 +1,57 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self,
+                 name: str,
+                 appetite: int = 7,
```
wQuelS, line 45: same

app/main.py
```diff
@@ -1 +1,57 @@
...
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+
+    def __init__(self,
+                 name: str,
+                 appetite: int = 3,
```
wQuelS, line 30: you don't need to pass default `appetite` param here, just pass it in `super`

wQuelS (changes requested):

app/main.py
YuriiKindrat: what about this?

app/main.py
YuriiKindrat: and this ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #352, opened 2023-02-10
Solution

viktoria-rybenchuk:


usrofgh (approved):
Good

app/main.py
```diff
@@ -1 +1,53 @@
...
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
```
wQuelS, line 16: `if self.is_hungry` already returns `boolean` value, so there is no need to write `if ... is True`

app/main.py
```diff
@@ -1 +1,53 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
```
wQuelS, line 18: `self.is_hungry = not self.is_hungry` would be better approach here

app/main.py
```diff
@@ -1 +1,53 @@
...
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
```
wQuelS, line 27: you don't have to pass default `appetite` here, since you have `super()` func

app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 appetite: int = 7,
```
wQuelS, line 40: same

app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([
+        animal.appetite for animal in animals if animal.feed()
+    ])
```
wQuelS, line 53: actually, `sum` can work with generators, so [] not needed here

wQuelS (changes requested):

app/main.py
```diff
@@ -1 +1,44 @@
...
+        super().__init__(name, 7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum(
+        animal.appetite for animal in animals if animal.feed()
```
YuriiKindrat, line 43: no need to make condition here, u can iterate through the animals and call feed method for every animal, it will return animal appetite

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #353, opened 2023-02-10
solution

usrofgh:


app/main.py
```diff
@@ -1 +1,50 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, 3, is_hungry)
```
ada-krav, line 29: It might be better to explicitly say what 3 is, to avoid a weird magic number in the code

app/main.py
```diff
@@ -1 +1,50 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, 7, is_hungry)
```
ada-krav, line 42: same here with 7

ada-krav (changes requested):

ada-krav (approved):

app/main.py
```diff
@@ -1 +1,50 @@
...
+    ) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
YuriiKindrat, line 50: no need to use [] brackets here, sum can work with generators

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #355, opened 2023-02-10
Solution

oigod:


app/main.py
```diff
@@ -1 +1,45 @@
...
+                 is_hungry: bool = True
+                 ) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: object) -> int:
+    return sum([animal.feed() for animal in animals])
```
Yulia-Mala, line 45: you can use generator here, without creating a list (just remove [])

Yulia-Mala (approved):
:)

app/main.py
oigod: you are right great thank you

wQuelS (approved):
Nice

PR #357, opened 2023-02-10
Solution

ada-krav:


app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
ramsty, line 18: You may remove else, as it said to avoid it in checklist

app/main.py
```diff
@@ -1 +1,41 @@
...
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        appetite = 3
+        super().__init__(name=name, appetite=appetite, is_hungry=is_hungry)
```
ramsty, line 25: No need to repeat yourself like "name=name", just name variables as they are in your parent class

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        appetite = 7
+        super().__init__(name=name, appetite=appetite, is_hungry=is_hungry)
```
ramsty, line 34: same here, also you can simplify your code by adding appetite into annotations or in super().__init__  directly

ramsty (changes requested):

app/main.py
ada-krav: I would like to disagree here, as adding it into annotation will allow user to change it while creating a Cat (or a Dog) change appetite while creating an instance. And there is a well-known fact that all cats eat 3 food points at a time :)

ramsty (approved):

app/main.py
```diff
@@ -1 +1,40 @@
...
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        appetite = 3
```
YuriiKindrat, line 23: no need to create variable for appetite

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #358, opened 2023-02-10
Solution

Anatolii-Poznyak:


oksana0410 (approved):

app/main.py
```diff
@@ -1 +1,51 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.name = name
+        self.is_hungry = is_hungry
```
YuriiKindrat, line 26: no need to rewrite this attributes here

app/main.py
```diff
@@ -1 +1,51 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.name = name
+        self.is_hungry = is_hungry
```
YuriiKindrat, line 38: the same

app/main.py
```diff
@@ -1 +1,51 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list[Animal]) -> int:
+    return sum(
+        animal.feed()
+        for animal in list_of_animals
+        if animal.is_hungry is True
```
YuriiKindrat, line 50: no need to make this check, it is already implemented in feed method

YuriiKindrat (changes requested):

app/main.py
Anatolii-Poznyak: roger that

YuriiKindrat (approved):

PR #363, opened 2023-02-10
Solution

VladyslavCherkashyn:


NazarHladaniuk (approved):

app/main.py
```diff
@@ -1 +1,46 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    points = 0
+    for el in animals:
+        points += el.feed()
+    return points
```
YuriiKindrat, line 46: use function sum here

YuriiKindrat: also el is a bad name for variable

YuriiKindrat (changes requested):

wQuelS (approved):
Nice

PR #364, opened 2023-02-10
Solution

ramsty:


app/main.py
```diff
@@ -1 +1,53 @@
...
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 3
+    ) -> None:
+        super(Cat, self).__init__(name, appetite, is_hungry)
```
NickSavchyn, line 30: super(Cat, self)
best practice is not to add parameters so super()

app/main.py
```diff
@@ -1 +1,53 @@
...
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 7
+    ) -> None:
+        super(Dog, self).__init__(name, appetite, is_hungry)
```
NickSavchyn, line 44: same with super

NickSavchyn (changes requested):

app/main.py
ramsty: But you know, my IDE(Pycharm) creates it by itself, so if it's good practice too(maybe not the best:)) I'd like to keep it not spending time to delete it.

app/main.py
ramsty: Same)

oksanaaam (approved):

Nattalli (approved):
Well done!

PR #368, opened 2023-02-11
'Solution'

alexlega:


app/main.py
```diff
@@ -1 +1,54 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_animals: list[Animal]) -> int:
+    total_food = 0
+    for animal in list_animals:
+        total_food += animal.feed()
+    return total_food
```
oksanaaam, line 54: You can use generator sum(...loop) for decreasing your code and writing in one line.

oksanaaam (changes requested):

app/main.py
alexlega: oke, thanks!

app/main.py
```diff
@@ -1 +1,51 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
wQuelS, line 21: `else` statement is redundant, since it will execute only if check in line16 is `False`

wQuelS (changes requested):

Nattalli (approved):
Well done!

PR #369, opened 2023-02-11
'Solution'

NazarHladaniuk:


app/main.py
```diff
@@ -1 +1,42 @@
...
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
```
dogmoony, line 16: you can delete "is True" it will still work.

app/main.py
```diff
@@ -1 +1,42 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = not self.is_hungry
```
dogmoony, line 18: its shorter to write "False" instead of not self.is_hungry

dogmoony (changes requested):

dogmoony (approved):

app/main.py
```diff
@@ -1 +1,42 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Dog, self).__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> int:
```
Nattalli, line 41: it's better to clarify what the type of list (list of integers, list of instances of some class, and so on) is inside of the type list. For example, list[int]

Nattalli (changes requested):

Nattalli (approved):
Good job!

PR #374, opened 2023-02-12
Solution

vladislav4i4eg:


app/main.py
```diff
@@ -1 +1,42 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    result = 0
+    for animal in animals:
+        result += animal.feed()
+    return result
```
Vasyl-Smutok, line 42: you can write it in one line 

HINT:
use `sum`

app/main.py
```diff
@@ -1 +1,42 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str,
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
```
Vasyl-Smutok, line 4:
```suggestion
    def __init__(
            self, 
            name: str,
            appetite: int,
            is_hungry: bool = True
    ) -> None:
```

Vasyl-Smutok (changes requested):

Vasyl-Smutok (approved):
Well Done 👍

PR #379, opened 2023-02-13
'Solution'

Bohdan-Kovalenko1992:


oksanaaam (approved):
good job

app/main.py
```diff
@@ -1 +1,54 @@
...
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, is_hungry)
+        self.appetite = appetite
```
YuriiKindrat, line 31: no need to rewrite attribute here, u can just pass a 3 ni super method

app/main.py
```diff
@@ -1 +1,54 @@
...
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 7,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, is_hungry)
+        self.appetite = appetite
```
YuriiKindrat, line 46: the same here

app/main.py
```diff
@@ -1 +1,54 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in list_of_animals])
```
YuriiKindrat, line 54: no need to use [] brackets here, sum can work with generators

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,52 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
```
wQuelS, line 27: no need to write it in init, just pass it right into the `super`. Fix everywhere

wQuelS (changes requested):

wQuelS (approved):
GJ

PR #380, opened 2023-02-13
Solution

destr34:


ProdyRodion (changes requested):

ProdyRodion (approved):
Good job!

app/main.py
```diff
@@ -1 +1,43 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
```
wQuelS, line 25: why two out of three args are keyword? Make them all keyword or positional. Fix everywhere

app/main.py
```diff
@@ -1 +1,43 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals])
```
wQuelS, line 43: no need in [], since `sum` can work with generators

wQuelS (changes requested):

PR #381, opened 2023-02-14
Solution

dogmoony:


app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    summary = 0
+    for animal in animals:
+        summary += animal.feed()
+    return summary
```
oksanaaam, line 46: you can better write generator sum(....loop)

oksanaaam (changes requested):

app/main.py
```diff
@@ -1 +1,43 @@
...
+            return 0
+        print(f"Eating {self.appetite} food points...")
+        self.is_hungry = False
+        return self.appetite
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
```
wQuelS, line 27: make all args kyeword or positional, don't mix. Fix everywhere

wQuelS (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #397, opened 2023-02-27
Solution

kostya-kononenko:


app/main.py
```diff
@@ -1 +1,49 @@
-#  write your code here
+class Animal:
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+        pass
```
katevasilenko, line 9: Please remove `pass`

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def print_name(self) -> str:
```
katevasilenko, line 11: This method return nothing. It only print

app/main.py
```diff
@@ -1 +1,49 @@
...
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True,
+                 appetite: int = 3) -> None:
+        super().__init__(name, appetite)
+        self.is_hungry = is_hungry
```
katevasilenko, line 29: Just put this to super too.

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def catch_mouse(self) -> str:
```
katevasilenko, line 31: Return nothing as well. Also switch this method to a staticmethod

app/main.py
```diff
@@ -1 +1,49 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
```
katevasilenko, line 38: The same as in Cat class

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def bring_slippers(self) -> str:
```
katevasilenko, line 40: Return nothing as well. Also switch this method to a staticmethod

app/main.py
```diff
@@ -1 +1,49 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list[object]) -> int:
+    sum_animal_appetite = 0
+    for animal in list_of_animals:
+        if Animal.feed(animal):
```
katevasilenko, line 47: You don’t need this to check

app/main.py
```diff
@@ -1 +1,49 @@
...
+            sum_animal_appetite += animal.appetite
+    return sum_animal_appetite
```
katevasilenko, line 49: Please use list comprehension here

katevasilenko (changes requested):

PR #401, opened 2023-03-08
Solution

V-Shkrobatskyi:


app/main.py
```diff
@@ -1 +1,50 @@
...
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        appetite = 7
```
Grayder0152, line 38: `
```suggestion
    def __init__(self, name: str, is_hungry: bool = True, appetite: int = 7) -> None:
        
```

app/main.py
```diff
@@ -1 +1,50 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            self.appetite = 0
+            return result
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        appetite = 3
```
Grayder0152, line 28:
```suggestion
    def __init__(self, name: str, is_hungry: bool = True,  appetite: int = 3) -> None:
       
```

app/main.py
```diff
@@ -1 +1,50 @@
-#  write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
+
+    def __init__(self, name: str, appetite: int, is_hungry: bool = True)\
```
Grayder0152, line 7: avoid using `\`

app/main.py
```diff
@@ -1 +1,50 @@
-#  write your code here
+from __future__ import annotations
+
+
+class Animal:
+    alive = []
```
Grayder0152, line 5: Where do you use this variable?

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,60 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    result = 0
+    for animal in animals:
+        result += animal.feed()
+    return result
```
Vasyl-Smutok, line 60: use `sum` and `comprehension`

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,57 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            result = self.appetite
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            self.appetite = 0
```
katevasilenko, line 23: You don’t need to set appetite to zero, it should be unchangeable. As a result, you don’t need 20 line

katevasilenko (changes requested):

app/main.py
```diff
@@ -1 +1,55 @@
...
+            appetite: int = 7
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Vasyl-Smutok, line 54:
```suggestion
def feed_animals(animals: list[Animal]) -> int:
```

Vasyl-Smutok (approved):
Great 👍

PR #410, opened 2023-03-23
Solution

goldenuni:


ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-#  write your code here
+from app.animal import Animal, Cat, Dog
+
+
+def feed_animals(ls_animals: list[Animal | Cat | Dog]) -> int:
```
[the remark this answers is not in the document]
goldenuni: @ArsenPidhoretskyi , yes but flake8 is arguing about that, cause Cat and Dog are not used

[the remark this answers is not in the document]
ArsenPidhoretskyi: Can you peen a screenshot?
Maybe because no need to import them here?

app/main.py
```diff
@@ -1 +1,5 @@
...
```
ArsenPidhoretskyi, line 4: also rename ls_animals to animals

app/main.py
```diff
@@ -1 +1,5 @@
-#  write your code here
+from app.animal import Animal, Cat, Dog
```
ArsenPidhoretskyi, line 1:
```suggestion
from app.animal import Animal```

File: app/main.py (hunk 2 above)
Comment (ArsenPidhoretskyi) [inline app/main.py:4]:
```suggestion
def feed_animals(ls_animals: list[Animal]) -> int:
```

app/main.py
```diff
@@ -1 +1,5 @@
...
+
+
+def feed_animals(ls_animals: list[Animal | Cat | Dog]) -> int:
+    return sum([animal.feed() for animal in ls_animals])
```
ArsenPidhoretskyi, line 5:
```suggestion
    return sum(animal.feed() for animal in ls_animals)
```
Converting to list is not needed

ArsenPidhoretskyi (changes requested):

app/main.py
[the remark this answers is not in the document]
goldenuni: image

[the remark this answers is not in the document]
goldenuni: it is happen, when I don`t import those classes

app/main.py
goldenuni: @ArsenPidhoretskyi 
image

Bohdan-Salamakha (approved):

PR #416, opened 2023-03-28
'Solution'

Maksym-Filipchuk:


app/main.py
```diff
@@ -1 +1,45 @@
...
+    @staticmethod
+    def bring_slippers() -> any:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    result = 0
+    for animal in animals:
+        result += animal.feed()
+    return result
```
ArsenPidhoretskyi, line 45: Use sum function

app/main.py
```diff
@@ -1 +1,45 @@
-#  write your code here
+class Animal:
+    animal_list = []
```
ArsenPidhoretskyi, line 2: Can you explain why you need this???

app/main.py
```diff
@@ -1 +1,45 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> any:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> any:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
ArsenPidhoretskyi, line 41: list of what?

ArsenPidhoretskyi (changes requested):
`any` is not accepted as type annotations! 
First of all any is function
Second you can provide better type annotations
Please read articles how you can add good type hinting

app/main.py
ArsenPidhoretskyi: This is not fixed

ArsenPidhoretskyi (changes requested):
Type annotations is not fixed also. Please provide correct type annotations, not `any`

app/main.py
```diff
@@ -1 +1,40 @@
...
+    return sum([animal.feed() for animal in animals])
```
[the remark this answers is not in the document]
Maksym-Filipchuk: print("The slippers delivered!")


what should I fix here ?
according to task, this function should recieve list 
or no ?

app/main.py
```diff
@@ -1 +1,40 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
katevasilenko, line 39:
```suggestion
def feed_animals(animals: list[Animal]) -> int:
```

katevasilenko (changes requested):

app/main.py
```diff
@@ -1 +1,40 @@
...
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
```
Grayder0152, line 23:
```suggestion
    def __init__(self, name: str, appetite:int = 3, is_hungry: bool = True) -> None:
        super().__init__(name, appetite=appetite, is_hungry=is_hungry)
```

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
```
Grayder0152, line 32: the same as above

Grayder0152 (changes requested):

PR #430, opened 2023-04-06
Solution

Fawnygod:


app/main.py
```diff
@@ -1 +1,53 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    appetite = 0
+    for animal in animals:
+        appetite += animal.feed()
+    return appetite
```
ArsenPidhoretskyi, line 53: use sum function

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+    ) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
ArsenPidhoretskyi, line 50: FYI, list is not needed
```suggestion
    return sum(animal.feed() for animal in animals)
```

ArsenPidhoretskyi (approved):
Well done!

PR #432, opened 2023-04-08
Solution last 08.04.2023

flatchers:
function feed_animals don't works, but in test successful.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,44 @@
-#  write your code here
+class Animal:
+    def __init__(
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True) \
+            -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> str:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,44 @@
...
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> str:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    result = 0
+    for i in animals:
+        result += i.feed()
```
ArsenPidhoretskyi, line 43: Use sum() function

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> str:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
ArsenPidhoretskyi, line 40: animals list of what?

app/main.py
```diff
@@ -1 +1,44 @@
-#  write your code here
+class Animal:
+    def __init__(
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True) \
+            -> None:
```
ArsenPidhoretskyi, line 7:
```suggestion
            is_hungry: bool = True
            ) -> None:
```

ArsenPidhoretskyi (changes requested):

app/main.py
flatchers: @ArsenPidhoretskyi numbers of function feed, no?

flatchers: @ArsenPidhoretskyi wow, I understood what you said

PR #436, opened 2023-04-12
Solution

ladno338:


app/main.py
```diff
@@ -1 +1,48 @@
...
+        else:
+            return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self, name: str, appetite: int = 3, is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, is_hungry)
+        self.appetite = appetite
```
ArsenPidhoretskyi, line 26: It works wrong. Due to your code
is_hungry will by always True
Please pass 3 parameters inside init as child class expect

ArsenPidhoretskyi: The same with dog

app/main.py
```diff
@@ -1 +1,48 @@
...
+            self, name: str, appetite: int = 7, is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, is_hungry)
+        self.appetite = appetite
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(ls: list) -> int:
```
ArsenPidhoretskyi, line 43: 1. accept list of what?
2. ls is not clear name. Rename to animals

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
...
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    all_eaten_points = 0
+    for i in animals:
+        eat = i.feed()
+        all_eaten_points += eat
+    return all_eaten_points
```
ArsenPidhoretskyi, line 48: Use sum function

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,46 @@
...
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+        self.appetite = appetite
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    all_eaten_points = 0
```
ArsenPidhoretskyi, line 44: This line is redundant. Remove it

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,44 @@
...
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self, name: str, appetite: int = 7, is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+        self.appetite = appetite
```
ArsenPidhoretskyi, line 37: This line is also not needed. Because it will be already set by super()

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self, name: str, appetite: int = 3, is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+        self.appetite = appetite
```
ArsenPidhoretskyi, line 26: This line is also not needed. Because it will be already set by super()

ArsenPidhoretskyi (changes requested):
Well done!

ArsenPidhoretskyi (approved):
Well done!

PR #446, opened 2023-04-23
Solution

markus2812:


app/main.py
```diff
@@ -1 +1,44 @@
-#  write your code here
+class Animal:
+    def __init__(
+            self, name: str,
+            appetite: int,
+            is_hungry: bool = True
+    ) -> None:
```
Vasyl-Smutok, line 6:
```suggestion
    def __init__(
            self, 
            name: str,
            appetite: int,
            is_hungry: bool = True
    ) -> None:
```

app/main.py
```diff
@@ -1 +1,44 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(ls_of_animals: list) -> int:
+    points = 0
+    for animal in ls_of_animals:
+        points += animal.feed()
+    return points
```
Vasyl-Smutok, line 44: try to use comprehension and `sum`

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,43 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(ls_of_animals: list) -> int:
```
ArsenPidhoretskyi, line 41: list of what? Provide inner type
also rename ls_of_animals to animals

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,43 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
ArsenPidhoretskyi, line 41: list of what?
Add inner type
list[AND HERE SOME TYPE]

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #448, opened 2023-04-25
Solution

jinjo0222988:


app/main.py
```diff
@@ -1 +1,8 @@
-#  write your code here
+from app.animals.animal import Animal
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    sum_of_food_points = 0
+    for animal in animals:
+        sum_of_food_points += animal.feed()
+    return sum_of_food_points
```
ArsenPidhoretskyi, line 8: Use sum function

ArsenPidhoretskyi (changes requested):
Also all classes must be inside main.py
Fix tests

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
Vasyl-Smutok, line 40: no need to create list 
```suggestion
    return sum(animal.feed() for animal in animals)
```

Vasyl-Smutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #449, opened 2023-04-27
Solution

Dimdimior:


app/main.py
```diff
@@ -1 +1,45 @@
...
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    count = 0
+    for animal in animals:
+        count += animal.appetite if animal.is_hungry else 0
+        animal.feed()
+    return count
```
ArsenPidhoretskyi, line 45: Use sum function. Also if is not needed here

app/main.py
```diff
@@ -1 +1,45 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+        self.name = name
+        self.is_hungry = is_hungry
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
ArsenPidhoretskyi, line 40: animals: list of what?
Provide inner type

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+        self.name = name
+        self.is_hungry = is_hungry
```
EdAlekseiev, line 34: You don't need to assign a name and is_hungry here, they will be assigned in `__init__` of Animal class.

app/main.py
```diff
@@ -1 +1,41 @@
...
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+        self.name = name
+        self.is_hungry = is_hungry
```
EdAlekseiev, line 24: You don't need to assign a name and is_hungry here, they will be assigned in `__init__` of Animal class.

EdAlekseiev (changes requested):

PR #452, opened 2023-04-29
'Solution'

YegorT:


app/main.py
```diff
@@ -1 +1,45 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
```
EdAlekseiev, line 29: Why is this method static?

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
```
EdAlekseiev, line 38: Why is this method static?

EdAlekseiev (changes requested):

app/main.py
YegorT: The `catch_mouse()` method in the `Cat` class is marked as static with `@staticmethod` decorator because it doesn't need to access or modify any instance variables or methods. 

A static method is bound to the class and not to the instance of the class, meaning that it can be called on the class itself, without the need for an instance of that class. In this case, `catch_mouse()` method does not rely on any instance-specific information (such as the cat's name, hunger level, etc.), and so it can be defined as a static method.

EdAlekseiev: Logically, this method must be called specifically on the instance of the class. If we call this method on class it will look illogically.

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_food_points = 0
+    for animal in animals:
+        total_food_points += animal.feed()
+    return total_food_points
```
Bohdan-Salamakha, line 43: do it using generator expression

Bohdan-Salamakha (changes requested):
v1

ArsenPidhoretskyi (approved):
Well done!

PR #463, opened 2023-05-06
Develop

denysbelei9:


app/main.py
```diff
@@ -1 +1,40 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([Animal.feed(animal) for animal in animals])
```
ArsenPidhoretskyi, line 40: Wrong call. You can call method to instance no need class

ArsenPidhoretskyi (changes requested):

PR #466, opened 2023-05-08
py-feed-animals

muriakula:


app/main.py
```diff
@@ -1 +1,46 @@
...
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, is_hungry)
+        self.appetite = 3
+        self.is_hungry = is_hungry
```
ArsenPidhoretskyi, line 24: You can pass this parameters inside init
```suggestion
        super().__init__(name, 3, is_hungry)
```

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, is_hungry)
```
ArsenPidhoretskyi, line 33: Not true. It will not work because due super call will be: 
name=name, appetite=is_hungry, is_hungry=True)
Fix this

app/main.py
```diff
@@ -1 +1,46 @@
...
+        self.appetite = 7
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
```
ArsenPidhoretskyi, line 41: list_of_animals rename to animals
also, specify the inner type for the list.

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Good job!

PR #467, opened 2023-05-08
'Solution'

SashaLee:


app/main.py
```diff
@@ -1 +1,40 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    @staticmethod
```
EdAlekseiev, line 25: Why is this method static? This must be python class level method

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
```
EdAlekseiev, line 34: Why is this method static? This must be python class level method

EdAlekseiev (approved):
Good job

PR #479, opened 2023-05-20
'Solution'

ilizh:


app/main.py
```diff
@@ -1 +1,40 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str, appetite: int,
+                 is_hungry: bool = True) -> None:
```
Bohdan-Salamakha, line 3:
```suggestion
    def __init__(
                 self, name: str,
                 appetite: int,
                 is_hungry: bool = True
                 ) -> None:
```

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+    def print_name(self) -> None:
+        print("Hello, I'm", self.name)
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print("Eating", self.appetite, "food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Bohdan-Salamakha, line 16: remove redundant "else"

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = 0
+    for animal in animals:
+        total_food_points += animal.feed()
+    return total_food_points
```
Bohdan-Salamakha, line 40: use generator expression here

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Bohdan-Salamakha, line 38: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

app/main.py
```diff
@@ -1 +1,40 @@
...
+    total_food_points = sum(animal.feed() for animal in animals)
+    return total_food_points
```
Bohdan-Salamakha, line 40: u don't need to create a redundant variable
```suggestion
    return sum(animal.feed() for animal in animals)
```

Bohdan-Salamakha (changes requested):

EdAlekseiev (changes requested):

PR #486, opened 2023-05-26
implemented solution

Oknametra:


app/main.py
```diff
@@ -1 +1,43 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+        self.name = name
+        self.is_hungry = is_hungry
```
ArsenPidhoretskyi, line 24: Redundant part of code. Super to the same! Read how super works
The same in Dog class

ArsenPidhoretskyi (changes requested):

app/main.py
Oknametra: yep, sry

vasylhnatiuk (approved):

PR #494, opened 2023-05-30
Solution

FadieievaA:


app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list[Union[Animal, Cat, Dog]]) -> int:
```
olena-hul, line 49:
```suggestion
def feed_animals(animal_list: list[Animal]) -> int:
```

Cat and Dog are subclasses of Animal, so we can just use Animal to annotate all its children

app/main.py
```diff
@@ -1 +1,53 @@
...
+    feed_sum = 0
+    for animal in animal_list:
+        feed_sum += animal.feed()
+    return feed_sum
```
olena-hul, line 53: Use `sum` function

olena-hul (changes requested):

FadieievaA:
виправлила

app/main.py
```diff
@@ -1 +1,47 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animal_list])
```
olena-hul, line 47:
```suggestion
    return sum(animal.feed() for animal in animal_list)
```

Won't reject because of this, but remember that we can use sum and other functions that work with iterables( for example `any` and `all`) in such a way too

PR #495, opened 2023-05-30
'Solution'

MarinaKhrunina:


app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum(Animal.feed(animal) for animal in animals)
```
ArsenPidhoretskyi, line 40: You can call method to instance no need to use class name

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #496, opened 2023-06-01
'Solution'

ant-komarov:


app/main.py
```diff
@@ -1 +1,49 @@
...
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
YuriiKindrat, line 26: why don't u pass this to super ?

app/main.py
```diff
@@ -1 +1,49 @@
...
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
```
YuriiKindrat, line 37: same here

app/main.py
```diff
@@ -1 +1,49 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list[Animal]) -> int:
+    points = 0
+    for animal in animals_list:
+        if animal.is_hungry:
+            points += animal.appetite
+            animal.feed()
+    return points
```
YuriiKindrat, line 49: why do u need to check animal.is_hungry this check already implemented in feed method
also thry to simplify this, using `sum`

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🍬

PR #497, opened 2023-06-01
Solution

volodymyr-komarnyckyi:


app/main.py
```diff
@@ -1 +1,41 @@
...
+        if not self.is_hungry:
+            return 0
+        print(f"Eating {self.appetite} food points...")
+        self.is_hungry = False
+        return self.appetite
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Cat, self).__init__(name, 3, is_hungry)
```
YuriiKindrat, line 24: After the last updates, there is no need to pass Cat, self to super(). It will take them automatically
the same is for Dog

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Well done!! 🌼

PR #498, opened 2023-06-01
Solution

AntonShpakovych:


app/main.py
```diff
@@ -1 +1,48 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: List[Animal]) -> int:
+    return sum([animal.feed() for animal in animals if animal.is_hungry])
```
LiudmylaKulinchenko, line 48: `sum` can work with generators, so [] are not needed

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if not self.is_hungry:
+            return 0
+
+        print(f"Eating {self.appetite} food points...")
+
```
LiudmylaKulinchenko, line 23: You can pass adding an empty line here

app/main.py
```diff
@@ -1 +1,48 @@
-#  write your code here
+from typing import List
+
+
+class Animal:
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 0,
```
LiudmylaKulinchenko, line 8: You should remove default value from `appetite`, there no animal that doesn't eat

LiudmylaKulinchenko (changes requested):

app/main.py
AntonShpakovych: good idea thx

YuriiKindrat (approved):

PR #499, opened 2023-06-01
Solution

diana-shyrokikh:


app/main.py
```diff
@@ -1 +1,54 @@
...
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            self.is_hungry = False
+
```
LiudmylaKulinchenko, line 18: You can skip adding an empty line here

app/main.py
```diff
@@ -1 +1,54 @@
...
+            print(f"Eating {self.appetite} food points...")
+
```
LiudmylaKulinchenko, line 20: You can even skip an empty line also here

app/main.py
```diff
@@ -1 +1,54 @@
...
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
```
LiudmylaKulinchenko, line 29: Cat\Dog classes should have two arguments - name, is_hungry

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):

PR #500, opened 2023-06-01
Solution

Tarasidze:
The solution has some trouble with default parameters in child classes. Separately these inheritors have undefined behavior)) However, all "pytest" tests have passed.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -15,15 +15,14 @@ class Animal:
             print(f"Eating {self.appetite} food points...")
             self.is_hungry = False
             return self.appetite
-        return int(0)
+        return 0
 
 
 class Cat(Animal):
     def __init__(self,
                  name: str,
-                 appetite: int = 3,
                  is_hungry: bool = True) -> None:
-        super().__init__(name, appetite, is_hungry)
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
 
     @staticmethod
     def catch_mouse() -> None:
@@ -33,9 +32,8 @@ class Cat(Animal):
 class Dog(Animal):
     def __init__(self,
                  name: str,
-                 appetite: int = 7,
                  is_hungry: bool = True) -> None:
-        super().__init__(name, appetite, is_hungry)
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
 
     @staticmethod
     def bring_slippers() -> None:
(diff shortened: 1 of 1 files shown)


Tarasidze:
The solution has some trouble with default parameters in child classes. Separately these inheritors have undefined behavior)) However, all "pytest" tests have passed.

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return int(0)
```
YuriiKindrat, line 18: no need to use int() here, u can just pass 0

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 appetite: int = 3,
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, appetite, is_hungry)
```
YuriiKindrat, line 26: image
fix this

app/main.py
```diff
@@ -1 +1,46 @@
...
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 appetite: int = 7,
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, appetite, is_hungry)
```
YuriiKindrat, line 38: same here

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals])
```
YuriiKindrat, line 46: sum can work with generators, so [] are not needed

YuriiKindrat (changes requested):

app/main.py
Tarasidze: Oh sorry! Forgot to fix it after my own tests ;)

app/main.py
Tarasidze: My bad)

LiudmylaKulinchenko (approved):
Good job 🎈

PR #501, opened 2023-06-01
Solution

Andriy-Sydorenko:


app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
LiudmylaKulinchenko, line 19: You don’t need `else` if you have a return statement in `if`

app/main.py
```diff
@@ -1 +1,47 @@
...
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list) -> int:
+    total_food_points = 0
+    for animal in animals_list:
```
LiudmylaKulinchenko, line 45: Try to rewrite in one line (use `sum`)

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):

PR #502, opened 2023-06-01
Develop

maccarets:


app/main.py
```diff
@@ -1 +1,48 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    result = 0
+    for animal in animals:
+        result += animal.feed()
+
+    return result
```
YuriiKindrat, line 48: could u write this in one line ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,45 @@
...
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+
+    return sum((animal.feed() for animal in animals))
```
LiudmylaKulinchenko, line 45: `sum` can work with generators, so () are not needed inside `sum()`

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🔥

PR #504, opened 2023-06-02
'Solution'

yakhlamova:


app/main.py
```diff
@@ -1 +1,38 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
YuriiKindrat, line 38: sum can work with generators, so [] are not needed

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Good job 🎈

PR #508, opened 2023-06-02
Solution

Gvilion:


app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(appetite=7, name=name, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals])
```
YuriiKindrat, line 40: `sum` can work with generators, so [] are not needed

YuriiKindrat: also write more descriptive type annotation for list

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #509, opened 2023-06-02
Solution

Kirontiko:


app/main.py
```diff
@@ -1 +1,41 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
LiudmylaKulinchenko, line 41: `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

app/main.py
Kirontiko: thx, I forgot about it:)

YuriiKindrat (approved):

PR #510, opened 2023-06-02
'Solution'

RedCatHot:


app/main.py
```diff
@@ -1 +1,44 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_points = 0
+    for animal in animals:
+        total_points += animal.feed()
+    return total_points
```
LiudmylaKulinchenko, line 44: Use sum() + list comprehention here instead

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🎈🧸

PR #511, opened 2023-06-02
Develop

IhorVoskoboinikov:


app/main.py
```diff
@@ -1 +1,55 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.appetite <= 0 or not self.is_hungry:
+            return 0
+        print(f"Eating {self.appetite} food points...")
+        eat_food = self.appetite
+        self.appetite -= self.appetite
```
YuriiKindrat, line 21: no need to do this, u should just return an appetite points, no need to change them

app/main.py
```diff
@@ -1 +1,55 @@
...
+        self.is_hungry = not self.is_hungry
+        return eat_food
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Cat, self).__init__(name=name, appetite=3)
```
YuriiKindrat, line 29: After the last updates, there is no need to pass Cat, self to super(). It will take them automatically

app/main.py
```diff
@@ -1 +1,55 @@
...
+        self.name = name
+        self.is_hungry = is_hungry
```
YuriiKindrat, line 31: no need to rewrite attributes here

app/main.py
```diff
@@ -1 +1,55 @@
...
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Dog, self).__init__(name=name, appetite=7)
+        self.name = name
+        self.is_hungry = is_hungry
```
YuriiKindrat, line 43: same is here

app/main.py
```diff
@@ -1 +1,55 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    all_food = 0
+    for animal in animals:
+        all_food += animal.feed()
+
+    return all_food
```
YuriiKindrat, line 55: try to simplify this using `sum`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,47 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    all_food = sum([animal.feed() for animal in animals])
```
LiudmylaKulinchenko, line 45: `sum` can work with generators, so [] are not needed

app/main.py
```diff
@@ -1 +1,47 @@
...
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if not self.is_hungry:
+            return 0
+        print(f"Eating {self.appetite} food points...")
+        self.is_hungry = not self.is_hungry
```
LiudmylaKulinchenko, line 20: Just set it to False here

LiudmylaKulinchenko (changes requested):

PR #513, opened 2023-06-02
Develop

imelnyk007:


app/main.py
```diff
@@ -1 +1,52 @@
...
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 3
+    ) -> None:
+        super(Cat, self).__init__(name, appetite, is_hungry)
```
LiudmylaKulinchenko, line 30: After the last updates, there is no need to pass Cat, self to super(). It will take them automatically

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 7
+    ) -> None:
+        super(Dog, self).__init__(name, appetite, is_hungry)
```
LiudmylaKulinchenko, line 44: Same here about passing Dog in the `super`

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,52 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 3
+    ) -> None:
```
YuriiKindrat, line 29: read the task again, init method should take 2 args
u can pass appetite directly to super method

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 7
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
```
YuriiKindrat, line 44: same is here

YuriiKindrat (changes requested):

PR #514, opened 2023-06-02
My Solution

kovaliskoveronika:


app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
LiudmylaKulinchenko, line 18: You don’t need `else` if you have a return statement in `if`

app/main.py
```diff
@@ -1 +1,46 @@
...
+                 is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    sum_points = 0
+    for animal in animals:
```

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):

PR #516, opened 2023-06-03
Solution

Kasatello:


app/main.py
```diff
@@ -1 +1,46 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Cat, self).__init__(name, 3, is_hungry)
```
LiudmylaKulinchenko, line 25: After the last updates, there is no need to pass Cat, self to super(). It will take them automatically

app/main.py
```diff
@@ -1 +1,46 @@
...
+        super(Dog, self).__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = 0
+    for animal in animals:
```

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):

PR #517, opened 2023-06-03
Solution

YevheniiIovchu:


app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    amount_of_feed = 0
+    for animal in animals:
+        amount_of_feed += animal.feed()
```
LiudmylaKulinchenko, line 46: Use `sum` function instead

LiudmylaKulinchenko: And mind that `sum` can work with generators, so [] are not needed

app/main.py
```diff
@@ -1 +1,51 @@
...
+    return amount_of_feed
+
+
+cat = Cat("", False)
+print(cat.is_hungry)
```
LiudmylaKulinchenko, line 51: Remove your local tests

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,44 @@
...
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
LiudmylaKulinchenko, line 44: `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

PR #518, opened 2023-06-04
Solution

YKonoplyov:


app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    food_eaten = 0
+
+    for animal in animals:
```
LiudmylaKulinchenko, line 42: Try to simplify (use `sum`)

LiudmylaKulinchenko (changes requested):

app/main.py
YKonoplyov: Understandable, thank you

YuriiKindrat (approved):

PR #519, opened 2023-06-04
Solution

Po-SvinSkipodla:


YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,51 @@
...
+    ) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals])
```
LiudmylaKulinchenko, line 51: `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

PR #520, opened 2023-06-04
Develop

12-deadblue:


app/main.py
```diff
@@ -1 +1,46 @@
...
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: List[Animal]) -> None:
+    points = 0
+    for animal in animals:
```

LiudmylaKulinchenko (changes requested):

app/main.py
12-deadblue: sum([animal.feed() for animal in animals]) так

YuriiKindrat (changes requested):

PR #522, opened 2023-06-05
my_commit

Mr-Furman:


app/main.py
```diff
@@ -1 +1,54 @@
...
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal: list[Animal]) -> int:
+    food_points = 0
+    for target in animal:
```
LiudmylaKulinchenko, line 52: Try to rewrite in one line (use `sum` and generator)

LiudmylaKulinchenko (changes requested):

app/main.py
Mr-Furman: It was done)

app/main.py
```diff
@@ -1 +1,51 @@
...
+        self.is_hungry = is_hungry
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal: list[Animal]) -> int:
+    return sum([target.feed() for target in animal])
```
LiudmylaKulinchenko, line 51: Good! Mind that `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

app/main.py
Mr-Furman: It was corrected

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🧸🎈🌺

PR #526, opened 2023-06-09
Implemented py-feed-animals

esemesenko:


app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Bohdan-Salamakha, line 42: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

app/main.py
```diff
@@ -1 +1,43 @@
...
+    return sum([animal.feed() for animal in animals])
```
Bohdan-Salamakha, line 43:
```suggestion
    return sum(animal.feed() for animal in animals)
```

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
```
EdAlekseiev, line 29: This method must be instance method

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
```
EdAlekseiev, line 39: This method must be instance method

EdAlekseiev (changes requested):

app/main.py
esemesenko: OK, but I don't understand why you can't use staticmethod?

YukihiroSM (approved):
Looks good. thanks for fixing!

PR #529, opened 2023-06-13
Solution

Veinmax:


app/main.py
```diff
@@ -1 +1,49 @@
...
+
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals if animal.is_hungry])
```
Grayder0152, line 49:
```suggestion
    return sum(animal.feed() for animal in animals if animal.is_hungry)
```

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = 0
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str,
+                 is_hungry: bool = True,
+                 appetite: int = 3) -> None:
```
ArsenPidhoretskyi, line 25:
```suggestion
    def __init__(
        self, 
        name: str,
        appetite: int = 3,
        is_hungry: bool = True,
   ) -> None:
```
Better to have the same order as the super. Also, check formatting fix the same in the Dog class

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Looks Good 👍

PR #541, opened 2023-06-28
Solution

pnakongit:


app/main.py
```diff
@@ -1 +1,48 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Cat | Dog]) -> int:
+    return sum(
+        [animal.feed()
+         for animal in animals]
+    )
```
vsmutok, line 48: no need to create list 
```suggestion
    return sum(animal.feed() for animal in animals)
```

pnakongit: Thx a lot. Noted

vsmutok (approved):
Well Done 👍

PR #548, opened 2023-07-02
Solution

Anastasia-Su:


app/main.py
```diff
@@ -1 +1,56 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list[Animal]) -> int:
+    food_points = 0
+    for animal in animal_list:
+        food_points += animal.feed()
+
+    return food_points
```
vsmutok, line 56: use `sum`  and comprehension

app/main.py
```diff
@@ -1 +1,56 @@
...
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if not self.is_hungry:
+            return 0
+        else:
```
vsmutok, line 18: remove redundant `else`

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,52 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 7) -> None:
+        super().__init__(name, appetite, is_hungry)
```
Grayder0152, line 44: the same as above

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 3
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
```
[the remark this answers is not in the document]
Anastasia-Su: Can't open this image, unfortunately. Could you please resend it?

ArsenPidhoretskyi, line 31: > Write Cat class which is a child of Animal. It should have the __init__ method with two arguments:
> 
> name - the name of a cat
> is_hungry - with True default value

It accepts two params, not 3. Then you can use:
Fix the same with Dog

app/main.py
```diff
@@ -1 +1,52 @@
...
+            is_hungry: bool = True,
+            appetite: int = 7) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list[Animal]) -> int:
```
ArsenPidhoretskyi, line 51: Rename animal_list to animals

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #552, opened 2023-07-06
solution of py-feed-animals/2

vkleshko:


app/main.py
```diff
@@ -1 +1,44 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
ArsenPidhoretskyi, line 20: else: not needed. Just  do return
```suggestion
        return 0
```

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_sum_of_appetite = 0
+    for animal in animals:
+        total_sum_of_appetite += animal.feed()
+    return total_sum_of_appetite
```
ArsenPidhoretskyi, line 44: Use sum function

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
ArsenPidhoretskyi, line 40: list of what? Provide inner type

ArsenPidhoretskyi (changes requested):

vkleshko:
i fixed all mistakes

PR #557, opened 2023-07-08
Solution

d-beschastnyi:


app/main.py
```diff
@@ -1 +1,51 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    result = 0
+    for animal in animals:
+        result += animal.feed()
+    return result
```
vsmutok, line 51: use `sum` and comprehension

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
...
+        super().__init__(name, is_hungry)
+        self.appetite = appetite
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum(Animal.feed(animal) for animal in animals)
```
vasylhnatiuk, line 48:
```suggestion
    return sum(animal.feed() for animal in animals)
```

vasylhnatiuk: don't use parent class if it not needed

app/main.py
```diff
@@ -1 +1,48 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 appetite: int = 3,
```
vasylhnatiuk, line 24: image
`Cat `init should not receive `appetite ` as an argument

vasylhnatiuk: the same with dog

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #561, opened 2023-07-10
created animals classes

vsafian:


app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.appetite and self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            points = self.appetite
+            self.appetite = 0
```
ArsenPidhoretskyi, line 18: Not said that we need to change it to 0

app/main.py
```diff
@@ -1 +1,48 @@
...
+                 ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.appetite and self.is_hungry:
```
ArsenPidhoretskyi, line 14:
```suggestion
        if self.is_hungry:
```
Not need to check by self.appetite

app/main.py
```diff
@@ -1 +1,48 @@
...
+            self.is_hungry = False
+            points = self.appetite
+            self.appetite = 0
+            return points
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True,
+                 appetite: int = 3) -> None:
```
ArsenPidhoretskyi, line 25:
```suggestion
    def __init__(
        self, name: str, appetite: int = 3, is_hungry: bool = True
    ) -> None:
```
No need to change order of params

app/main.py
```diff
@@ -1 +1,48 @@
...
+                         appetite=appetite)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True,
+                 appetite: int = 7) -> None:
```
ArsenPidhoretskyi, line 37:
```suggestion
    def __init__(
        self, name: str, appetite: int = 7, is_hungry: bool = True
    ) -> None:
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str,
```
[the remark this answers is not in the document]
vasylhnatiuk: the same with Dog

vasylhnatiuk (approved):

PR #562, opened 2023-07-12
solution

ivanzh13:


app/main.py
```diff
@@ -1 +1,50 @@
...
+            appetite=7,
+            is_hungry=is_hungry
+        )
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list) -> int:
```
ArsenPidhoretskyi, line 49: 1. rename animal_list to animals
2. add inner type to annotation. list[Animal]

app/main.py
```diff
@@ -1 +1,50 @@
...
+    return sum([animal.feed() for animal in animal_list])
```
ArsenPidhoretskyi, line 50:
```suggestion
    return sum(animal.feed() for animal in animal_list)
```
Convert to list is redundant

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
Changes requested

app/main.py
ivanzh13: Please explain how to use: list[Animal] or list[Animal | Cat | Dog ]

ArsenPidhoretskyi: Better to use list[Animal] it already means that will work with all child classes

vasylhnatiuk (approved):

PR #569, opened 2023-07-21
Solution

AllaKuksa:


app/main.py
```diff
@@ -1 +1,43 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
ArsenPidhoretskyi, line 23: else: is not needed
```suggestion
        return 0
```

app/main.py
```diff
@@ -1 +1,43 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum(Animal.feed(animal) for animal in animals)
```
ArsenPidhoretskyi, line 43: Call methods to instance not to class

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #583, opened 2023-08-02
Dog and Cat that inherits from Animal and returns its appetite.

IProskurnytskyi:


app/main.py
```diff
@@ -1 +1,45 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        self.appetite = 0
+        return self.appetite
```
pavlejviki, line 21: this method should return 0, not self.appetite, in case is_hungry is False.

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    for animal in animals:
+        animal.feed()
+    return sum([animal.appetite for animal in animals])
```
pavlejviki, line 45:
```suggestion
    return sum(animal.feed() for animal in animals)
```

app/main.py
```diff
@@ -1 +1,45 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, is_hungry=is_hungry, appetite=7)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    for animal in animals:
```
pavlejviki, line 43: You can simplify this method to have just one line of code.

pavlejviki (changes requested):

Dimosphen1 (approved):
LGTM, nice work!

PR #586, opened 2023-08-09
Solution

nyanplague:


app/main.py
```diff
@@ -1 +1,41 @@
-#  write your code here
+class Animal:
+    def __init__(self,
```
Kroll410, line 2: Pass the paramters to the nextline https://stackoverflow.com/a/41591509

Kroll410 (changes requested):

app/main.py
```diff
@@ -1 +1,41 @@
...
+                 name: str,
```
Kroll410, line 3: Pass the paramters to the nextline https://stackoverflow.com/a/41591509

Kroll410 (changes requested):

app/main.py
nyanplague: @Kroll410 Pushed the update 👍

IvanRamyk (approved):

PR #587, opened 2023-08-09
'Solution'

lgSilay:


app/main.py
```diff
@@ -1 +1,52 @@
-#  write your code here
+from __future__ import annotations
+
+
+class Animal:
+    def __init__(self,
```
Kroll410, line 5: Pass the paramters indents like in [this](https://stackoverflow.com/a/41591509) example

Kroll410 (changes requested):
Overall is good

app/main.py
```diff
@@ -1 +1,54 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    sum_of_food = 0
+    for animal in list_of_animals:
+        if animal.is_hungry:
```
IvanRamyk, line 51: use `feed` method to avoid logic duplication

IvanRamyk (changes requested):

app/main.py
lgSilay: Okey, i fix this)

app/main.py
```diff
@@ -1 +1,54 @@
...
+        super().__init__(name, 7)
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    sum_of_food = 0
```
IvanRamyk, line 49: could be implemented by list comprehension :) 

but it's not critical

IvanRamyk (approved):

PR #588, opened 2023-08-09
Solution

eduardhabryd:


app/main.py
```diff
@@ -1 +1,57 @@
-#  write your code here
+from typing import List
+
+
+class Animal:
+    def __init__(self,
```
Kroll410, line 5: Pass there variables with an indent like [here ](https://stackoverflow.com/questions/24988162/define-functions-with-too-many-arguments-to-abide-by-pep8-standard)

Kroll410 (changes requested):
Overall is good, just apply changes to all function indents

IvanRamyk (approved):

PR #593, opened 2023-08-10
Implemented class Animal and its 2 subclasses, as well feed_animals function.

KristianYe:


app/main.py
```diff
@@ -1 +1,48 @@
-#  write your code here
+class Animal:
+    def __init__(self,
```
IvanRamyk, line 2: please, use black to fix indentation

IvanRamyk (changes requested):

Kroll410 (approved):

PR #595, opened 2023-08-10
Solution py-feed-animals

katryana:


app/main.py
```diff
@@ -1 +1,56 @@
...
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self,
```
IvanRamyk, line 40: please, use black to fix indentation

app/main.py
```diff
@@ -1 +1,56 @@
...
+                 ) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(hungry_animals: list[Animal]) -> int:
+    food_points = 0
```
IvanRamyk, line 53: sum could be found using list comprehension

IvanRamyk (changes requested):

PR #596, opened 2023-08-11
solution implemented

maxkatkalov:


app/main.py
```diff
@@ -1 +1,62 @@
...
+            name: str,
+            appetite: int,
+            is_hungry: bool = True
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(
+            self
```
IvanRamyk, line 13: here you can define a method in one line

IvanRamyk: and I believe it's better choice

app/main.py
```diff
@@ -1 +1,62 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+    ) -> None:
+        super().__init__(
```
IvanRamyk, line 50: do you actually multiple lines here

app/main.py
```diff
@@ -1 +1,62 @@
...
+            7,
+            is_hungry
+        )
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(lst: list[Animal]) -> int:
```
IvanRamyk, line 61: lst? please, improve naming

IvanRamyk (changes requested):

Kroll410 (approved):

PR #606, opened 2023-08-11
Solution

anna-lybid:


app/main.py
```diff
@@ -1 +1,45 @@
...
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    points = 0
+    for animal in animals:
+        if animal.is_hungry is True:
```
IvanRamyk, line 42:
```suggestion
        if animal.is_hungry:
```
if you check a bool value you don't need to compare it explicitly

IvanRamyk: Use `feed` method here to avoid logic duplication

app/main.py
```diff
@@ -1 +1,45 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    points = 0
```
IvanRamyk, line 40: could be calculated using list comprehension

app/main.py
```diff
@@ -1 +1,45 @@
-#  write your code here
+class Animal:
+
+    def __init__(self,
```
IvanRamyk, line 3: fix indentation

IvanRamyk (changes requested):

app/main.py
anna-lybid: > 

Thanks.

app/main.py
Kroll410: +

app/main.py
```diff
@@ -1 +1,39 @@
...
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
```
Kroll410, line 14: We can't use `is` keyword with operation upon values (is_hungry: bool), because `is` addresses to the memory. Use not or `==` instead

Kroll410 (changes requested):

PR #622, opened 2023-08-18
solution py-feed-animals

IvDyriv:


app/main.py
```diff
@@ -1 +1,47 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str,
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
```
olena-hul, line 4: Place each argument in a separate line

olena-hul (changes requested):

app/main.py
IvDyriv: Okay, thank you)

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = sum(animal.feed()
+                            for animal in animals if animal.is_hungry)
+    return total_food_points
```
Bohdan-Salamakha, line 48: simplify it
```suggestion
    return  sum(
    animal.feed() for animal in animals if animal.is_hungry
    )
```

Bohdan-Salamakha (changes requested):

app/main.py
IvDyriv: Cool, thank you.

app/main.py
```diff
@@ -1 +1,48 @@
...
+                 is_hungry: bool = True) -> None:
+
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Bohdan-Salamakha, line 45: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #634, opened 2023-08-25
Implemented feed_animals

osharlai:


app/main.py
```diff
@@ -1 +1,62 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    feed_result = 0
+    for animal in animals:
+        if animal.is_hungry:
+            feed_result += animal.appetite
+            animal.feed()
```
pavlejviki, line 60: You can simplify this method using sum() function , comprehension and taking into account that feed returns self.appetite and checks if the animal is hungry

pavlejviki (changes requested):

app/main.py
osharlai: Done, code has become more simplify, as for me - it become more difficult to understand..

app/main.py
```diff
@@ -1 +1,56 @@
...
+    ) -> Dog:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals if animal.is_hungry])
```
Bohdan-Salamakha, line 56:
```suggestion
    return sum(animal.feed() for animal in animals if animal.is_hungry)
```

Bohdan-Salamakha (changes requested):

app/main.py
osharlai: Oo, fixed

Polyakiv-Andrey (approved):
Good job!

PR #635, opened 2023-08-27
Solution

prochigor:


app/main.py
Daniil-Pankieiev, line 1: Great

Daniil-Pankieiev (approved):

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(hungry_animals: list) -> int:
+    return sum(
+        feed_all_animals.feed()
+        for feed_all_animals in hungry_animals
```
LiudmylaKulinchenko, line 47: You should consider to rename variables
Method feed_animals gets not only hungry_animals. Thats why you have `if not self.is_hungry:` check in feed animals
Also in the loop you iterate by _animal_, so `feed_all_animals` is a bad name here

LiudmylaKulinchenko (changes requested):

app/main.py
prochigor: I see, I can get all animals and feed hungry animals

app/main.py
```diff
@@ -1 +1,48 @@
...
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum(
+        hungry_animals.feed()
```
LiudmylaKulinchenko, line 46: It will be just animal and method `feed` will check if it's hungry
So you could use `for animal in animals`

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum(
+        animal.feed()
+        for animal in animals
+    )
```
LiudmylaKulinchenko, line 48: No need to split into lines if it fits in one

LiudmylaKulinchenko (approved):
Nice! 😎

app/main.py
prochigor: Yes, earlier it didn't fit)

PR #646, opened 2023-08-31
Solution: Feed animals.

lobodInI:


app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    result = 0
+    for animal in animals:
+        result += Animal.feed(animal)
```
pavlejviki, line 47: You can use sum() and comprehension here, and call the feed method from the instance of the Animal class, not from the class itself

pavlejviki (changes requested):

app/main.py
lobodInI: 👍

Bohdan-Salamakha (approved):

PR #648, opened 2023-09-01
Fed the animals)

sberdianskyi:
Wrote animal classes and implemented the feed_animals function as requested

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,44 @@
-#  write your code here
+class Animal:
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True
+                 ) -> None:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,44 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list[Animal]) -> int:
+    food_points = 0
+    for animal in animals_list:
+        food_points += animal.feed()
+    return food_points
```
Grayder0152, line 44: or
```suggestion
    return sum(animal.feed() for animal in animals_list)

```

Grayder0152 (approved):

PR #649, opened 2023-09-01
Solution-py-feed-animals

Nadiia-developer:


app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
pavlejviki, line 21: else is redundant

app/main.py
```diff
@@ -1 +1,51 @@
...
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = sum(
+        [animal.feed() for animal in animals if animal.is_hungry]
+    )
```
pavlejviki, line 50: You can make this more efficient by using a generator expression. Generator expressions are similar to list comprehensions, but they don't create the whole list at once. Instead, they generate each value on the fly. You can achieve this simply by removing the square brackets

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #650, opened 2023-09-01
Solutin

AndriiYe:


app/main.py
```diff
@@ -1 +1,56 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=0, is_hungry=True)
```
pavlejviki, line 30:
```suggestion
        super().__init__(name, appetite=3, is_hungry=is_hungry)
```

app/main.py
```diff
@@ -1 +1,56 @@
...
+        self.is_hungry = is_hungry
+        self.appetite = 3
```
pavlejviki, line 32: just pass the correct arguments when calling super()

app/main.py
```diff
@@ -1 +1,56 @@
...
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=0, is_hungry=True)
```
pavlejviki, line 42: see comment above

app/main.py
```diff
@@ -1 +1,56 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_feed_points = 0
+    for animal in animals:
+        total_feed_points += Animal.feed(animal)
+    return total_feed_points
```
pavlejviki, line 56: you can simplify this function using sum and a generator expression. call feed() method from the instance of the class, not from the class itself

pavlejviki (changes requested):

AndriiYe:
Thanks for the advice now it looks math better

app/main.py
```diff
@@ -1 +1,48 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum((Animal.feed(animal) for animal in animals))
```
Bohdan-Salamakha, line 48:
```suggestion
    return sum(Animal.feed(animal) for animal in animals)
```

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Bohdan-Salamakha, line 47: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]

Bohdan-Salamakha (changes requested):

AndriiYe:
got it

Bohdan-Salamakha (approved):

PR #654, opened 2023-09-03
Solution

Manynik:


app/main.py
```diff
@@ -1 +1,52 @@
...
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum(animal.feed() for animal in animals if animal.is_hungry)
```
Dimosphen1, line 52: Delete the `if animal.is_hungry` condition from `sum()`, this logic is already implemented in the `feed` method.

app/main.py
```diff
@@ -1 +1,52 @@
...
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 7,
+            is_hungry: bool = True
```
Dimosphen1, line 42: You should pass only `name` and `is_hungry` arguments to the `__init__` here according to the task.

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
```
Dimosphen1, line 30: You should pass only `name` and `is_hungry` arguments to the `__init__` here according to the task.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
```
Dimosphen1, line 42: In case you use keyword arguments, it would be preferable to call `__init__`  with `name=name` as well to be consistent here and on line 29.

Dimosphen1 (approved):

PR #657, opened 2023-09-06
Solution

SerhiiBudko:


app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Dimosphen1, line 20: Do not use `else` here, code will not execute after the `return` statement anyway.

app/main.py
```diff
@@ -1 +1,44 @@
...
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
```
Dimosphen1, line 26: In case you pass keyword arguments in the `__init__`, it would be preferable to pass `name=name` as well to maintain the same structure here.

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def catch_mouse(self) -> print():
```
Dimosphen1, line 28: You should not state that the return type would be `print` with a call, it would be `None` because there is no `return` statement here.

app/main.py
```diff
@@ -1 +1,44 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> print():
```
Dimosphen1, line 36: You should not state that the return type would be `print` with a call, it would be `None` because there is no `return` statement here.

Dimosphen1 (changes requested):

SerhiiBudko:
Job done as per job-description)

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    total_food_point = 0
+    for animal in list_of_animals:
+        total_food_point += animal.feed()
+    return total_food_point
```
Dimosphen1, line 43: You could use `sum()` here and write this code in one line.

Dimosphen1 (approved):

app/main.py
SerhiiBudko: As per yours instruction, the code was realised via method sum() in this way: 
total_food_point = sum(animal.feed() for animal in list_of_animals)
return total_food_point

PR #661, opened 2023-09-13
Solution

Depaton:


app/main.py
```diff
@@ -1 +1,51 @@
...
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    sum_of_appetite = 0
+    for animal in animals:
```
Polyakiv-Andrey, line 47: use list generator and function sum()

app/main.py
```diff
@@ -1 +1,51 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        upd_appetite = self.appetite
+        if self.is_hungry is True:
+            self.is_hungry = not self.is_hungry
+            print(f"Eating {self.appetite} food points...")
+            self.appetite = 0
+        else:
```
Polyakiv-Andrey, line 22: else should be omitted

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        upd_appetite = self.appetite
+        if self.is_hungry is True:
+            self.is_hungry = not self.is_hungry
+            print(f"Eating {self.appetite} food points...")
+            self.appetite = 0
+            return upd_appetite
+        return 0
```
Dimosphen1, line 23: There is no need to redefine the `self.appetite`, you should return its value or 0 here. Rewrite your code to encounter that.

app/main.py
```diff
@@ -1 +1,50 @@
...
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        upd_appetite = self.appetite
+        if self.is_hungry is True:
```
Dimosphen1, line 18: There is no need to explicitly check that the boolean `True` is stored there. Remove the `is True` part from your if condition.

app/main.py
```diff
@@ -1 +1,50 @@
...
+            self.is_hungry = not self.is_hungry
+            print(f"Eating {self.appetite} food points...")
+            self.appetite = 0
+            return upd_appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
```
Dimosphen1, line 28: In case you pass keyword arguments, add` name=name` here and on line 37 as well. It is better to stick to one approach in passing arguments.

app/main.py
```diff
@@ -1 +1,50 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    sum_of_appetite = sum(
+        [animal.appetite for animal in animals if animal.is_hungry is True]
+    )
+    for animal in animals:
+        animal.feed()
+    return sum_of_appetite
```
Dimosphen1, line 50: You should write a generator expression inside of the `sum` where you will call the `feed` method for each `Animal` instance in the `animals` list. Write this method in one line, and do not use the `[]` brackets inside of the `sum`.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,44 @@
...
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    sum_of_appetite = sum(animal.feed() for animal in animals)
+    return sum_of_appetite
```
Dimosphen1, line 44: Do not allocate variable `sum_of_appetite` here, return the result of the calculation directly.

app/main.py
Dimosphen1: This comment was not implemented. Additional comment is left.

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
```
Dimosphen1, line 17: Write your code in the following way
```suggestion
        if self.is_hungry:
```

Dimosphen1 (changes requested):
Several minor changes were requested.

viktoria-rybenchuk (approved):
Nice

PR #665, opened 2023-09-18
'Solution_feed-animals'

vvg04021968:


app/main.py
```diff
@@ -1 +1,52 @@
...
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
```
Dimosphen1, line 27: Your `Cat` and `Dog` classes should not have an `appetite` argument in the `__init__` according to the task condition. Remove it and pass a default value in the `super().__init__` instead.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #670, opened 2023-09-20
Solution

MatsalakO:


app/main.py
```diff
@@ -1 +1,44 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
viktoria-rybenchuk, line 21: Don't use an extra else statement.

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
```
viktoria-rybenchuk, line 26: In case you pass keyword arguments in the __init__, it would be preferable to pass name=name as well to maintain the same structure here.

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list) -> int:
+    count_feed = 0
+    for animal in animal_list:
+        count_feed += animal.feed()
+    return count_feed
```
viktoria-rybenchuk, line 44: make it in one line with `sum`

viktoria-rybenchuk (changes requested):

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #671, opened 2023-09-21
Implemented Feed animals.

AndriiHomeniuk1:


app/main.py
```diff
@@ -1 +1,53 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: List[Animal]) -> int:
+    total_food_points = 0
+    for animal in animals:
+        if animal.is_hungry:
+            total_food_points += animal.feed()
+
+    return total_food_points
```
ArsenPidhoretskyi, line 53: Use sum function

ArsenPidhoretskyi (changes requested):

app/main.py
AndriiHomeniuk1: Fixed it

Yurnerroo (approved):

PR #673, opened 2023-09-25
Solution

loplicat:


app/main.py
```diff
@@ -1 +1,56 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    food_points = 0
+    for animal in animals:
+        food_points += animal.feed()
+    return food_points
```
Bohdan-Salamakha, line 56: use generator expression here

Bohdan-Salamakha (changes requested):

app/main.py
loplicat: Thanks for checking. I already added it.

Dimosphen1 (approved):
LGTM, well done!

PR #675, opened 2023-09-27
solution

Familenko:


app/main.py
```diff
@@ -1 +1,50 @@
-#  write your code here
+class Animal:
+
+    def __init__(
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True) -> None:
```
LiudmylaKulinchenko, line 7: Fix formatting: brackets must be located in one of two options: open and start with the text or have a line break between the text

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+
+    required_food = 0
+
+    for animal in animals:
```

app/main.py
```diff
@@ -1 +1,50 @@
...
+            return self.appetite
+
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(appetite=3, name=name, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
```
LiudmylaKulinchenko, line 29: U can use staticmethod here

app/main.py
```diff
@@ -1 +1,50 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(appetite=7, name=name, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
```
LiudmylaKulinchenko, line 37: same here

LiudmylaKulinchenko (changes requested):

app/main.py
LiudmylaKulinchenko: Method feed returns animal.appetite, use it. Also there is a check if animal is hungry in the feed method, remove it here
Rewrite whole function in one line

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🐇

PR #678, opened 2023-09-28
'Solution'

lylakivan:


app/main.py
```diff
@@ -1 +1,52 @@
...
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_food_points = 0
+    for animal in animals:
+        food_points = animal.feed()
+        total_food_points += food_points
+    return total_food_points
```
Polyakiv-Andrey, line 52: use list generator and function max()

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    food_points_needed = [animal.feed() for animal in animals
+                          if animal.is_hungry]
+    total_food_points = sum(food_points_needed)
+    return total_food_points
```
viktoria-rybenchuk, line 51: try to do it in one line

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,50 @@
...
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    food_points_needed = [animal.feed() for animal in animals
+                          if animal.is_hungry]
+    return sum(food_points_needed)
```
Dimosphen1, line 50: Place the generator expression inside of the `sum`, there is no need to create a list via list comprehension here. In this way, you could omit to create the `food_points_needed` variable. Besides that, there is no need for the if condition here, the check is already performed in the `feed` method.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
...
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
viktoria-rybenchuk, line 48: use a generator expression,  which is more memory-efficient  instead of a list comprehension

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (changes requested):
You didn't fix it. Read this https://stackoverflow.com/questions/47789/generator-expressions-vs-list-comprehensions

pavlejviki (approved):
Good job!

PR #683, opened 2023-10-03
Solution

dkibalenko:


app/main.py
```diff
@@ -1 +1,96 @@
...
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name=name, appetite=3)
+        self.is_hungry = is_hungry
```
pavlejviki, line 63: you should pass is_hungry to super().__init__ instead  of reassigning its value

app/main.py
```diff
@@ -1 +1,96 @@
...
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, is_hungry)
+        if is_hungry:
```
pavlejviki, line 81: follow the same logic as in Cat class 

pass appetite to super().__init__

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,99 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super_init = partial(super().__init__, name=name, is_hungry=is_hungry)
```
pavlejviki, line 82:
```suggestion
super().__init__(name, appetite=7, is_hungry=is_hungry)
```

pavlejviki: that's it

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #684, opened 2023-10-04
Solution

vshvanska:


app/main.py
```diff
@@ -1 +1,49 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    food_points = 0
+    for animal in animals:
+        food_points += animal.feed()
+    return food_points
```
Naz-iv, line 49: As per checklist, we should use list comprehension instead of for loop, try to rewrite it. Also there is way to avoid creating temporary variable food_points if you resolve task with list comprehension.

Naz-iv (changes requested):

app/main.py
```diff
@@ -1 +1,46 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str,
+                 appetite: int = 3,
+                 is_hungry: bool = True) -> None:
```
LiudmylaKulinchenko, line 25: Cat\Dog classes should have two arguments - name, is_hungry

app/main.py
```diff
@@ -1 +1,46 @@
...
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
LiudmylaKulinchenko, line 46: `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🧸

PR #686, opened 2023-10-05
'Solution'

romanshupak:


app/main.py
```diff
@@ -1 +1,50 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: [object]) -> int:
```
ArsenPidhoretskyi, line 46: [object] it is not good)
Use list[And here some inner type]

ArsenPidhoretskyi: Also inside use sum() function

ArsenPidhoretskyi (approved):
Well done!

PR #687, opened 2023-10-05
Solution

D1st3f:


app/main.py
```diff
@@ -1 +1,44 @@
...
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
LiudmylaKulinchenko, line 44: `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice! 😎

PR #691, opened 2023-10-07
Solution

petrykivd:


YuliaHladyshkevych (approved):

app/main.py
```diff
@@ -1 +1,42 @@
...
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
```
LiudmylaKulinchenko, line 17: If you use backward condition, this method could look more pretty, as for me. Optional, but you could give a try

LiudmylaKulinchenko (changes requested):

PR #693, opened 2023-10-07
Solution

DenPrislipskyi:


skoval7 (approved):

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(Animal, appetite=3)
+        self.name = name
+        self.is_hungry = is_hungry
+
+    def catch_mouse(self) -> None:
```
LiudmylaKulinchenko, line 27: You can use staticmethod, self is not used here

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(Animal, appetite=7)
+        self.name = name
+        self.is_hungry = is_hungry
+
+    def bring_slippers(self) -> None:
```
LiudmylaKulinchenko, line 38: same here

app/main.py
```diff
@@ -1 +1,48 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list) -> int:
+    total_appetite = 0
+    for animal in animal_list:
```

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice! 😎

PR #694, opened 2023-10-07
Solution

kostomeister:


app/main.py
```diff
@@ -1 +1,57 @@
...
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
LiudmylaKulinchenko, line 35: Just pass it in super

app/main.py
```diff
@@ -1 +1,57 @@
...
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
```
LiudmylaKulinchenko, line 49: same here

app/main.py
```diff
@@ -1 +1,57 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: List[Animal]) -> int:
+    return sum([animal.feed() for animal in animals_list])
```
LiudmylaKulinchenko, line 57: `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,55 @@
...
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: List[Animal]) -> int:
```
LiudmylaKulinchenko, line 54: You can use list instead of typing.List for annotations from python 3.9

LiudmylaKulinchenko (approved):
Nice! 😎

app/main.py
kostomeister: got it!

PR #695, opened 2023-10-08
Created the Animal class and two following subclasses and a function

raychw:
Created the Animal class and two following subclasses and a function

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,58 @@
-#  write your code here
+class Animal:
+    animals = []
+
+    def __init__(
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True
+    ) -> None:
+
+        self.name = name
+        self.is_hungry = is_hungry
+        self.appetite = appetite
+        self.animals.append(self)
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        food_points = 0
+
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            food_points += self.appetite
+
+            return food_points
+
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: Animal.animals) -> int:
+    total_food_points = 0
+
+    for animal in animals:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,58 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: Animal.animals) -> int:
+    total_food_points = 0
+
+    for animal in animals:
+        total_food_points += animal.feed()
+
+    return total_food_points
```
Dimosphen1, line 58: Write this method in one line with the use of `sum`. Pass generation expression there, not list comprehension.

app/main.py
```diff
@@ -1 +1,58 @@
...
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
```
Dimosphen1, line 44: You should pass `is_hungry` in the `__init__` as well, add this for `Dog` and `Cat`.

app/main.py
```diff
@@ -1 +1,58 @@
...
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: Animal.animals) -> int:
```
Dimosphen1, line 52: You should explicitly indicate here that you expect to obtain a list of `Animal` class instances for `animals`.

app/main.py
```diff
@@ -1 +1,58 @@
...
+
+    def feed(self) -> int:
+        food_points = 0
+
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            food_points += self.appetite
+
+            return food_points
```
Dimosphen1, line 27: There is no need for the `food_points` variable, return `self.appetite` here directly.

app/main.py
```diff
@@ -1 +1,58 @@
...
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True
+    ) -> None:
+
+        self.name = name
+        self.is_hungry = is_hungry
+        self.appetite = appetite
+        self.animals.append(self)
```
Dimosphen1, line 14: `animals` is a class attribute, call it from the `Animal` class, not from `self`.

Dimosphen1 (changes requested):

app/main.py
raychw: image

but it contradicts the condition, though

raychw: figured it out

app/main.py
```diff
@@ -1 +1,48 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: [Animal]) -> int:
```
Dimosphen1, line 47: It would be better to explicitly indicate that you expect to obtain a list for the `animals` variable, e.g
. 
```suggestion
def feed_animals(animals: list[Animal]) -> int:
```

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #697, opened 2023-10-08
Solution

mrve4nick:


app/main.py
```diff
@@ -1 +1,48 @@
...
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    points_sum = 0
+    for animal in animals:
+        if animal.is_hungry:
+            points_sum += animal.appetite
+            animal.feed()
+
+    return points_sum
```
Dimosphen1, line 48: Write this method in one line using `sum`. Pass generator expression there, not list comprehension.

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
```
Dimosphen1, line 37: Make this method static, self is not used inside of it.

app/main.py
```diff
@@ -1 +1,48 @@
...
+            return self.appetite
+
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    def catch_mouse(self) -> None:
```
Dimosphen1, line 29: Make this method static, self is not used inside of it.

Dimosphen1 (changes requested):

app/main.py
mrve4nick: > Write this method in one line using `sum`. Pass generator expression there, not list comprehension.

what does generator expression have to do with it?

Dimosphen1: You should pass a generator expression inside of `sum` to obtain the result of the `feed` operation for animals.

Dimosphen1 (changes requested):
Improvements to the `feed_animals` function were requested.

app/main.py
mrve4nick: > You should pass a generator expression inside of `sum` to obtain the result of the `feed` operation for animals.

I misunderstood you, thought I needed somehow to yield appetite points in the sum function😅

app/main.py
```diff
@@ -1 +1,44 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum(animal.feed() for animal in animals if animal.is_hungry)
```
Dimosphen1, line 44: There is no need to add if condition here, the check will be performed inside of the `feed` method. Remove it.

Dimosphen1 (changes requested):
One minor change was requested.

Dimosphen1 (approved):
LGTM, well done!

PR #698, opened 2023-10-08
Solution

sakhaline:


RVChornyy (commented):
I suggest you to use @staticmethod for catch_mouse and bring_slippers.
In both methods attribute "self" is not beeing used.

RVChornyy (approved):

mykolamateichuk (approved):

app/main.py
```diff
@@ -1 +1,43 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list["Animal", "Dog", "Cat"]) -> int:
```
LiudmylaKulinchenko, line 42: No need to use quotes
Also no need to specify Dog and Cat, they are anyway Animal instances, because those classes inherits from Animal

LiudmylaKulinchenko (approved):
Nice! 😎

PR #699, opened 2023-10-08
Feeding animals

SelFisher1488:


RVChornyy (approved):
Good!

katyakoziar (approved):

app/main.py
```diff
@@ -1 +1,44 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
LiudmylaKulinchenko, line 21: no need to use `else` if you have return in your if statement

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    food_points_needed = 0
+    for animal in animals:
+        food_points_needed += animal.feed()
+    return food_points_needed
```

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great! 🚀

PR #702, opened 2023-10-08
Solution

OlehShumov:


SevKrok (approved):

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: List[Animal]) -> int:
+    total_food_points = 0
+    for animal in animal_list:
+        food_points = animal.feed()
+        total_food_points += food_points
```
LiudmylaKulinchenko, line 44: Try to rewrite in one line (use `sum`)

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: List[Animal]) -> int:
```
LiudmylaKulinchenko, line 40: From python 3.9 you can use list for annotations instead of typing.List

LiudmylaKulinchenko (approved):
Nice! 😎

PR #704, opened 2023-10-09
Solution

katyakoziar:


app/main.py
```diff
@@ -1 +1,44 @@
...
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
```
RVChornyy, line 28: most probably you should use @staticmethod, because attribute "self" is not beeing used.

app/main.py
```diff
@@ -1 +1,44 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
```
RVChornyy, line 36: the same here

RVChornyy (changes requested):

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
LiudmylaKulinchenko, line 20: You don’t need `else` if you have a return statement in `if`

app/main.py
```diff
@@ -1 +1,46 @@
...
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    food_points = 0
+    for animal in animals:
```

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice! 😎

PR #705, opened 2023-10-09
Solution

Wonsky1:


RabenkoYevhenii (approved):
Everything seems to be ok, but in my opinion you have to write what are the numbers you call in super(). It is better to write `super().__init__(name, appetite=n, is_hungry=is_hungry)`, where n is 3 for cats and 7 for dogs

app/main.py
```diff
@@ -1 +1,43 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: [Animal]) -> int:
```
LiudmylaKulinchenko, line 42: With type annotation like this your function will except any collection with Animal instances
It will work fine, but in the task description you have the next: ` Implement feed_animals function which takes a list of animals.` So you should specify that function gets **list** of animals

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🎈

PR #706, opened 2023-10-09
Solution

roffi37:


app/main.py
```diff
@@ -1 +1,47 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    sum_of_food_points = 0
```
Shatrovskyi, line 44: In my opinion, it's better to use list comprehension with sum(). It's going to be more readable.

Shatrovskyi (approved):

app/main.py
```diff
@@ -1 +1,47 @@
...
+    for animal in animals:
```

app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def print_name(self) -> str:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            self.is_hungry = False
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        if not self.is_hungry:
```
LiudmylaKulinchenko, line 21: It's same as `else`. But in this case you don't even need `else`, because you have a return statement in `if`

LiudmylaKulinchenko (changes requested):

Lyutillis (approved):

LiudmylaKulinchenko (approved):
Great! ☀

PR #708, opened 2023-10-09
Solution

RabenkoYevhenii:


app/main.py
```diff
@@ -1 +1,44 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
fedorov-s-od, line 39: You can write a 1-liner for this function )

app/main.py
```diff
@@ -1 +1,44 @@
...
+        if self.is_hungry is True:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
```
fedorov-s-od, line 23: In my opinion, positional arguments are redundant here and in the __init__ method in the class "Dog"

Lyutillis (approved):

app/main.py
LiudmylaKulinchenko: As for me, it's fine, especially for appetite
But probably you should also add for name so here will be consistent style

app/main.py
LiudmylaKulinchenko: Agree. Try to rewrite in one line, use `sum`

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Keep it up 🔥

PR #714, opened 2023-10-09
Solution Feed animals

Oxbay:


app/main.py
```diff
@@ -1 +1,41 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals])
```
LiudmylaKulinchenko, line 41: `sum` can work with generators, so [] are not needed

app/main.py
```diff
@@ -1 +1,41 @@
-#  write your code here
+class Animal:
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True
+                 ) -> None:
```
LiudmylaKulinchenko, line 6: Fix formatting: brackets must be located in one of two options: open and start with the text or have a line break between the text

LiudmylaKulinchenko (changes requested):

app/main.py
Oxbay: Yep

app/main.py
Oxbay: ok, done)

LiudmylaKulinchenko (approved):
Nice! 🚀

PR #716, opened 2023-10-10
Solution

mishagitcode:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        else:
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
```
EdAlekseiev, line 33: Why is the method static?

app/main.py
```diff
@@ -1 +1,47 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
```
EdAlekseiev, line 41: Why is method static?

EdAlekseiev (changes requested):

app/main.py
mishagitcode: Because these methods just print info. And PyCharm recommended me two options: 1) to move the methods from classes or 2) to make it static

EdAlekseiev (approved):
Good job

PR #720, opened 2023-10-11
'Solution'

Illya-Maznitskiy:


app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: [Animal]) -> int:
+    food_points = sum([animal.feed()
+                       for animal in animals_list
+                       if animal.is_hungry])
```
Dimosphen1, line 44: Do not use list comprehension inside of `sum`, pass generator expression instead. Besides that, you do not need to perform the if condition check here, it is already performed inside of the `feed` method.

app/main.py
```diff
@@ -1 +1,45 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: [Animal]) -> int:
```
Dimosphen1, line 41: It is a better practice to be more concrete and define that you expect to obtain a list of Animal instances in type annotation
```suggestion
def feed_animals(animals_list: list[Animal]) -> int:
```

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,43 @@
...
+            is_hungry: bool = True
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> any:
```
Polyakiv-Andrey, line 15: return type annotation is inncorect

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1 +1,43 @@
...
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals_list: list[Animal]) -> int:
+    food_points = sum(animal.feed() for animal in animals_list)
+    return food_points
```
Dimosphen1, line 43: There is no need for the `food_points` variable, you could omit using it and return the `sum` directly in the future.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #722, opened 2023-10-13
Solution

ibulakevych:


app/main.py
```diff
@@ -1 +1,54 @@
...
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True
+                 ) -> None:
+        super().__init__(name, is_hungry=True, appetite=3)
+        self.name = name
+        self.is_hungry = is_hungry
+
+    @staticmethod
```
EdAlekseiev, line 31: Why is this method static?  Method can be static if it is not part of the instance.

app/main.py
```diff
@@ -1 +1,54 @@
...
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True
+                 ) -> None:
+        super().__init__(name, is_hungry=True, appetite=7)
+        self.name = name
+        self.is_hungry = is_hungry
+
+    @staticmethod
```
EdAlekseiev, line 45: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #726, opened 2023-10-15
Added solution

IlyaGrynyshyn:


app/main.py
```diff
@@ -1 +1,41 @@
...
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, is_hungry)
+        self.appetite = 7
```
pavlejviki, line 34: you can pass the value for self.appetite straight to super().__init__ in the same way as in class Cat

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,40 @@
...
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
```
viktoria-rybenchuk, line 28: Why do you need 'self' here? This method is static. Read checklist

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
```
viktoria-rybenchuk, line 36: According to the comment above, why do you need 'self' here?

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+
+def feed_animals(animals: list) -> int:
```
viktoria-rybenchuk, line 39: specify type annotation for animals

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #727, opened 2023-10-15
Solution

hbilous:


app/main.py
```diff
@@ -1 +1,42 @@
...
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals])
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.name = name
+        super().__init__(name, appetite=3)
+        self.is_hungry = is_hungry
```
ArsenPidhoretskyi, line 29: You can pass 3 parameters into super method

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+def feed_animals(animals: list) -> int:
```
ArsenPidhoretskyi, line 21: animals list of what? Add inner type

ArsenPidhoretskyi (changes requested):

app/main.py
EdAlekseiev: Still not fixed

app/main.py
```diff
@@ -1 +1,38 @@
...
```
EdAlekseiev, line 21:
```suggestion
def feed_animals(animals: list[Animal]) -> int:
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #729, opened 2023-10-17
Solution

paashkovaaa:


app/main.py
```diff
@@ -1 +1,42 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
```
EdAlekseiev, line 29: Why is this method static? A method can be static if it is not part of the instance.

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
```
EdAlekseiev, line 36: The same comment as above

EdAlekseiev (changes requested):

app/main.py
paashkovaaa: PyCharm is providing two weak warnings, suggesting that these methods may be static. Also I have 2 errors "Method must have a first parameter, usually called 'self'". But I dont need it here.

NikOneZ1 (approved):
Well done 👍

PR #730, opened 2023-10-18
Solution

Ded31:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True) -> None:
+        super().__init__(name=name, is_hungry=is_hungry, appetite=3)
+
+    @staticmethod
```
EdAlekseiev, line 27: Why is this method static? A method can be static if it is not part of the instance.

app/main.py
```diff
@@ -1 +1,47 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True) -> None:
+        super().__init__(name=name, is_hungry=is_hungry, appetite=7)
+
+    @staticmethod
```
EdAlekseiev, line 38: The same comment as above

app/main.py
```diff
@@ -1 +1,47 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    food_points = 0
+    for animal in animals:
+        food_points += animal.feed()
```
EdAlekseiev, line 46: You can use method `sum` here.
https://www.w3schools.com/python/ref_func_sum.asp

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #734, opened 2023-10-19
Sulution

okien1:


app/main.py
```diff
@@ -1 +1,52 @@
...
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
viktoria-rybenchuk, line 52:
```suggestion
    return sum(animal.feed() for animal in animals)
```

viktoria-rybenchuk: use generator instead of list comprehension

viktoria-rybenchuk (changes requested):

PR #735, opened 2023-10-20
Solution

IvanMozhar:


app/main.py
```diff
@@ -1 +1,56 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list[Animal]) -> int:
+    total_appetite = 0
+    for animal in list_of_animals:
+        if animal.feed():
+            total_appetite += animal.appetite
+    return total_appetite
```
pavlejviki, line 56: use sum() function with a generator expression to simplify the code

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
pavlejviki, line 20: you can remove else, as its redundant

pavlejviki (approved):
Well done, just make a minor fix

PR #740, opened 2023-10-24
Solution

MykytaPD09:


app/main.py
```diff
@@ -1 +1,64 @@
...
+
+    def bring_slippers(self) -> str:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = 0
+    for animal in animals:
+        food_points = animal.feed()
+        total_food_points += food_points
```
pavlejviki, line 62: use sum() function and a generator expression to simplify the code

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,59 @@
...
+    def catch_mouse(self) -> str:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry:
+            bool = True
```
Dimosphen1, line 46: Place `bool = True` on the same line with `is_hungry` to match its type annotation. You could indicate in your type annotation that this argument is optional as well throughout your code where it is used.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,61 @@
...
+            name: str,
+            is_hungry: Optional[bool] = True
+    ) -> None:
+        super().__init__(
+            name,
+            appetite=7,
+            is_hungry=is_hungry
+        )
+
+    def bring_slippers(self) -> str:
```
Dimosphen1, line 56: You could use the `staticmethod` decorator for methods that do not use `self`.

Dimosphen1 (approved):

PR #746, opened 2023-10-28
solution

Andrey1104:


app/main.py
```diff
@@ -1 +1,61 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    result = []
+    for animal in animals:
+        if isinstance(animal, Animal) and animal.is_hungry is True:
+            animal.feed()
+            result.append(animal.appetite)
```
viktoria-rybenchuk, line 60: This operation can be simplified into a one-liner using generator expression with `sum` for improved code conciseness.

app/main.py
```diff
@@ -1 +1,61 @@
-#  write your code here
+class Animal:
+
```
viktoria-rybenchuk, line 2: Remove the blank line between the class definition and the `__init__` method it  follows common Python coding conventions

viktoria-rybenchuk: Fix in all places

viktoria-rybenchuk (changes requested):

app/main.py
Andrey1104: i tryed, but i have no idea, how in generator execute 2 functions (animal.(feed() and generate list with animal.appetite), explain to me, please.

viktoria-rybenchuk: Instead of initializing an empty list to collect results, we use a generator expression inside sum.

The generator expression iterates over each animal in the animals list.

If the conditions are met, the animal.feed() method is called directly within the generator expression.

The generator expression itself produces a sequence of appetite values returned by animal.feed().

sum then calculates the total sum of these values and returns it.

viktoria-rybenchuk (changes requested):

app/main.py
Andrey1104: ok, maybe i find solution)))

Andrey1104: aaaa, self.feed return self.appetite)))) I forgot about this. maybe my last request not so correct)

app/main.py
```diff
@@ -1 +1,58 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum(
+        [animal.feed()
+         for animal in animals
+         if isinstance(animal, Animal)
+         and animal.is_hungry is True]
```
viktoria-rybenchuk, line 57: Instead of creating a list and then summing, you can directly sum the results)

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,58 @@
...
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
```
Dimosphen1, line 27: For the `Cat` and `Dog` classes you should take only two arguments `name` and `is_hungry`, you should pass the `appetite` with value specific to the class directly to the `super().__init__` statement and do not leave an opportunity to change the `appetite`.

app/main.py
```diff
@@ -1 +1,58 @@
...
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, is_hungry)
+        self.appetite = appetite
```
Dimosphen1, line 31: Pass `appetite` to the `super().__init__` and do not redefine it out of this statement.

app/main.py
```diff
@@ -1 +1,58 @@
...
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 7,
+            is_hungry: bool = True) -> None:
```
Dimosphen1, line 43: Place `) -> None:` on a new line to match the code style when you do not have arguments on the first line with an opening bracket.

app/main.py
```diff
@@ -1 +1,58 @@
...
+        self.appetite = appetite
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum(
+        [animal.feed()
```
Dimosphen1: There is no need to make a list comprehension inside of `sum`, create a generator expression instead and pass it to the `sum`. Besides that, checks that go after `if` are redundant, you could omit them because you initially expect that you will receive only `Animal` instances, and logic that checks if the animal is hungry or not is placed inside of the `.feed()` method.

app/main.py
```diff
@@ -1 +1,58 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum(
+        animal.feed()
+         for animal in animals
+         if isinstance(animal, Animal)
+         and animal.is_hungry is True
```
Dimosphen1: Checks that go after `if` are redundant, you could omit them because you initially expect that you will receive only `Animal` instances, and logic that checks if the animal is hungry or not is placed inside of the `.feed()` method.

app/main.py
```diff
@@ -1 +1,58 @@
...
+            is_hungry: bool = True) -> None:
+        super().__init__(name, is_hungry)
+        self.appetite = appetite
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Dimosphen1, line 52: You could be more specific here indicating that you expect to obtain a list of `Animal` instances for animals.

app/main.py
```diff
@@ -1 +1,58 @@
...
+    return sum(
+        animal.feed()
+        for animal in animals
+        if isinstance(animal, Animal)
+        and animal.is_hungry is True
```
Dimosphen1, line 57: There is no need to perform checks that go after `if` because you initially expect to obtain a list of `Animal` instances and the logic that checks whether the animal is hungry is placed inside of the `.feed()` method.

Dimosphen1 (changes requested):

Dimosphen1 (changes requested):
Changes from the previous comments were not implemented.

app/main.py
Andrey1104: when i pass apeetite  in super, i get uninspected argumen. please, explain how it work?

Andrey1104: how i must pass apeetite in super?

app/main.py
```diff
@@ -1 +1,54 @@
...
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+            ) -> None:
```
Dimosphen1, line 41: Correct the indentation level for this line.

Dimosphen1 (changes requested):
Several minor changes were requested to fix the flake8 error.

Dimosphen1 (approved):
LGTM, well done!

Andrey1104:
what mean LGTM ?

PR #748, opened 2023-10-31
Solution

SashaTarnavsky:


app/main.py
```diff
@@ -1 +1,47 @@
...
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.name = name
+        self.is_hungry = is_hungry
```
YukihiroSM, line 28: super().__init__ will do the same, why do you need to do this?

app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def catch_mouse(self) -> str:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.name = name
+        self.is_hungry = is_hungry
```
YukihiroSM, line 39: super().init will do the same, why do you need to do this?

YukihiroSM (changes requested):

app/main.py
SashaTarnavsky: you are right! thanks

app/main.py
```diff
@@ -1 +1,43 @@
...
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Cat, self).__init__(name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> str:
+        print("The hunt began!")
```
NikOneZ1, line 30: This method could be static

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Dog, self).__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> str:
+        print("The slippers delivered!")
```
NikOneZ1, line 39: This method could be static

NikOneZ1 (changes requested):

app/main.py
```diff
@@ -1 +1,45 @@
...
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Cat, self).__init__(name, appetite=3, is_hungry=is_hungry)
+
+    @staticmethod
```
EdAlekseiev, line 29: Why is this method static?

app/main.py
```diff
@@ -1 +1,45 @@
...
+    def catch_mouse() -> str:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Dog, self).__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
```
EdAlekseiev, line 39: Why is this method static?

EdAlekseiev (changes requested):

app/main.py
SashaTarnavsky: the catch_mouse() and bring_slippers() method does not use any attributes or instance methods of the Cat and Dog class. Therefore, in theory, these methods can be declared as static.

EdAlekseiev: A static method is a method that belongs to a class rather than an instance of a class. But, this method describes behaviour of an instance of the class. So, this method can't be static

EdAlekseiev (changes requested):

app/main.py
SashaTarnavsky: I understand, thank you. I will fix the code now

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super(Dog, self).__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> str:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum(animal.feed() for animal in animals if animal.is_hungry)
```
ArsenPidhoretskyi, line 43:
```suggestion
    return sum(animal.feed() for animal in animals)
```
If statement is redundant

ArsenPidhoretskyi (changes requested):

app/main.py
SashaTarnavsky: please check my code

app/main.py
SashaTarnavsky: thanks, fixed

YukihiroSM (approved):
Good job!

SashaTarnavsky:
> Good job!

thanks for the help!

PR #749, opened 2023-10-31
Solution

Taran21V:


app/main.py
```diff
@@ -1 +1,48 @@
...
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True
+                 ) -> None:
+        super().__init__(name, 3, is_hungry)
+        self.is_hungry = is_hungry
+
+    @staticmethod
```
EdAlekseiev, line 30: Why is this method static? A static method is a method that belongs to a class rather than an instance of a class. However, your method contains behaviour of an instance of a class.

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True
+                 ) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
```
EdAlekseiev, line 42: The same comment as above

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,46 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True
+                 ) -> None:
+        super().__init__(name, 3, is_hungry)
+        self.is_hungry = is_hungry
```
ArsenPidhoretskyi, line 28: This is redundant line

app/main.py
```diff
@@ -1 +1,46 @@
...
+                 name: str,
+                 is_hungry: bool = True
+                 ) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_animals: list) -> int:
```
ArsenPidhoretskyi, line 45:
```suggestion
def feed_animals(animals: list[Animal]) -> int:
```
list_ prefix is redundant. Also need to add inner type annotations

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #755, opened 2023-11-03
Solution

aleksandra-prognimak:


app/main.py
```diff
@@ -1 +1,50 @@
...
+
+
+def feed_animals(animals: list) -> int:
+    return sum(
+        [
+            animal.appetite and animal.feed()
+            for animal in animals
+            if animal.is_hungry
+        ]
+    )
```
LiudmylaKulinchenko, line 50: sum works with generators so [] are not needed

LiudmylaKulinchenko (approved):
Keep it up 🌺

PR #760, opened 2023-11-06
Solution

eLQeR:


app/main.py
```diff
@@ -1 +1,64 @@
...
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 3
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
```
EdAlekseiev, line 42: Why is this method static? A static method is a method that belongs to a class rather than an instance of a class. However, this method contains logic for instance

app/main.py
```diff
@@ -1 +1,64 @@
...
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 7,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
```
EdAlekseiev, line 54: The same comment as above

app/main.py
```diff
@@ -1 +1,64 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_food_points = 0
+    for animal in animals:
+        if animal.is_hungry:
+            total_food_points += animal.feed()
+    return total_food_points
```
EdAlekseiev, line 64: You can simplify this method with using `sum` function

EdAlekseiev (changes requested):

PR #764, opened 2023-11-09
Solution for py_feed_animals

VladyslavBazhyn:


app/main.py
```diff
@@ -1 +1,58 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    feed_points_sum = 0
+
+    for animal in [animal for animal in animals
+                   if animal.is_hungry]:
```
EdAlekseiev, line 54: You don't need to make additional check `if animal.is_hungry`, because you make it inside `feed`. Also, you can rewrite this function to one line with using function `sum`

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #765, opened 2023-11-10
solution)

xborismenx:


app/main.py
```diff
@@ -1 +1,46 @@
...
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+
+        self.is_hungry = is_hungry
+
+    @staticmethod
```
EdAlekseiev, line 28: Why is this method static? A static method is a method that belongs to a class rather than an instance of a class. However, this method contains logic for instance

app/main.py
```diff
@@ -1 +1,46 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+
+        self.is_hungry = is_hungry
+
+    @staticmethod
```
EdAlekseiev, line 39: The same comment as above

app/main.py
```diff
@@ -1 +1,46 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
EdAlekseiev, line 19: `else` block is redundant
```suggestion
    def feed(self) -> int:
        if self.is_hungry is True:
            print(f"Eating {self.appetite} food points...")
            self.is_hungry = False
            return self.appetite
        return 0
```

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+
+        self.is_hungry = is_hungry
```
EdAlekseiev, line 26: Pass this argument to `super` method
```suggestion
        super().__init__(name, appetite=3,  is_hungry= is_hungry)
```

app/main.py
```diff
@@ -1 +1,46 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+
+        self.is_hungry = is_hungry
```
EdAlekseiev, line 37: The same comment as above

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    all_food_point = sum([animal.feed() for animal in animals])
+    return all_food_point
```
EdAlekseiev, line 46:
```suggestion
def feed_animals(animals: list) -> int:
    return sum([animal.feed() for animal in animals])
```

EdAlekseiev (changes requested):

YukihiroSM (approved):
Nicely done

PR #766, opened 2023-11-11
Solution

ShcherbynaSv:


app/main.py
```diff
@@ -1 +1,60 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    food = 0
+    for animal in animals:
+        if animal.is_hungry:
+            animal.feed()
+            food += animal.appetite
+    return food
```
viktoria-rybenchuk, line 60: use a one-liner with a generator expression and the `sum()` function.

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,55 @@
...
+        return self.appetite
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=3)
```
Dimosphen1, line 32: You should pass the `is_hungry` to the `super().__init__(` as well, the user can define this value during the creation of the `Cat` or `Dog` instance. Add it to the `Dog` and `Cat` classes.

app/main.py
```diff
@@ -1 +1,55 @@
...
+        self.name = name
+        self.is_hungry = is_hungry
```
Dimosphen1, line 34: There is no need to redefine the `self.name` and `self.is_hungry` after the call of `super`, the `__inti__` of the `Animal` class does all the job. Remove these lines from `Dog` and `Cat` classes.

app/main.py
```diff
@@ -1 +1,55 @@
...
+        super().__init__(name, appetite=7)
+        self.name = name
+        self.is_hungry = is_hungry
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum(animal.feed() for animal in animals if animal.is_hungry)
```
Dimosphen1, line 55: There is no need to check if the animal is hungry here, this logic is implemented in the `feed()` method. Remove it from this line.

Dimosphen1 (changes requested):

app/main.py
ShcherbynaSv: Hello!
When I add is_hungry to the super().__init__ method in Cat and Dog classes I get a mistake that init takes two positional arguments, but three were giwen.

image

Dimosphen1: In your screenshot, you removed the `is_hungry` from the `__init__` of the `Dog` and `Cat` classes (what is wrong) and added the hardcoded value to the `super().__init__(...` (what is wrong as well). Because of that, you obtain the error. Return the `is_hungry: bool = True` so that you can initialize the `Cat` and `Dog` instances with two arguments as it was before and pass the `is_hungry` value to the  `super().__init__(...` without hardcoding.
```suggestion
        super().__init__(name, appetite=3, is_hungry=is_hungry)
```

Dimosphen1 (changes requested):

ShcherbynaSv:
Thanks. Now it works. :)

main.py
```diff
@@ -0,0 +1,51 @@
...
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
```
viktoria-rybenchuk, line 35: Do you use `self` here? Use a @staticmethod for some functions. Read checklist

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (changes requested):
you have duplicated `main.py`. Please remove one of them)

PR #767, opened 2023-11-12
Solution

Vityk-4W:


app/main.py
```diff
@@ -1 +1,96 @@
-#  write your code here
+from __future__ import annotations
```
ArsenPidhoretskyi, line 1: this import is redundant

ArsenPidhoretskyi (changes requested):
tests are not passed, remove import

Vityk-4W:
Deteted redundant import (from __future__ import annotations).
Deleted doc strings in Cat/Dog classes.

ArsenPidhoretskyi (approved):
Well done!

PR #769, opened 2023-11-14
Develop

Moonralex:


app/main.py
```diff
@@ -1 +1,59 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
viktoria-rybenchuk, line 20: `else` is redundant

app/main.py
```diff
@@ -1 +1,59 @@
...
+            is_hungry=is_hungry
+        )
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animal_list])
```
viktoria-rybenchuk, line 59:
```suggestion
    return sum(animal.feed() for animal in animal_list)
```

viktoria-rybenchuk: this variant is preferred for its improved memory efficiency, lazy evaluation, and cleaner syntax. It is a good practice to use generator expressions when the result is only needed for iteration or when applying functions like sum.

viktoria-rybenchuk (changes requested):

app/main.py
Moonralex: I didn't think about that, thank you. It looks better

viktoria-rybenchuk (approved):
Good job!

PR #770, opened 2023-11-16
solution

OleksiiMi:


app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
vasylhnatiuk, line 24: else is useless here

vasylhnatiuk (changes requested):

app/main.py
OleksiiMi: return = 0, if I delete this, feed_animals() write:  TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'

OleksiiMi: I understand, thank you

vasylhnatiuk (approved):

PR #784, opened 2023-11-25
ready solution!

job-artem:


app/main.py
```diff
@@ -1 +1,49 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
ArsenPidhoretskyi, line 21:
```suggestion
       return 0
```

app/main.py
```diff
@@ -1 +1,49 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    sum_feed = 0
+    for animal in animals:
+        sum_feed += animal.feed()
+    return sum_feed
```
ArsenPidhoretskyi, line 49: Use sum function

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 7,
```
Dimosphen1, line 35: You should not accept appetite in the `__init__` for `Dog`, the value should be hardcoded as it is done in the `Cat` class according to the task condition.

app/main.py
```diff
@@ -1 +1,45 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    def catch_mouse(self) -> None:
```
Dimosphen1, line 27: You could use `staticmethod` decorator for methods that do not utilize `self`.

app/main.py
```diff
@@ -1 +1,45 @@
...
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
```
Dimosphen1, line 24: It is a better practice to place all arguments inside of the `__init__` on a new line as it is done in other classes.

app/main.py
```diff
@@ -1 +1,45 @@
-#  write your code here
+class Animal:
+    def __init__(
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True
```
Dimosphen1, line 6: You could indicate that arguments that have default value are optional in a type annotation using the following syntax:
```suggestion
            is_hungry: bool | None = True
```

Dimosphen1 (changes requested):

PR #787, opened 2023-11-28
Feed animals

Ananiev-Vitalii:


app/main.py
```diff
@@ -1 +1,50 @@
...
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
```
EdAlekseiev, line 15:
```suggestion
        if self.is_hungry:
```

app/main.py
```diff
@@ -1 +1,50 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
```
EdAlekseiev, line 29: Why is this method static? A static method is a method that belongs to a class rather than an instance of a class. However, this method contains logic for instance

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+        self.name = name
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
```
EdAlekseiev, line 40: The same comment as above

app/main.py
```diff
@@ -1 +1,50 @@
...
+
+
+def feed_animals(animals: list) -> int:
+    food_points = sum(
+        animal.appetite
+        for animal in animals if animal.is_hungry)
+    for animal in animals:
+        if animal.is_hungry:
+            animal.feed()
```
EdAlekseiev, line 49: You can simplify this method
```suggestion
    food_points = sum(
        animal.feed()
        for animal in animals)
    return food_points
```

EdAlekseiev (changes requested):

PR #797, opened 2023-12-08
Develop

OlgaRozd:


app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = 0
+    for animal in animals:
+        total_food_points += animal.feed()
+    return total_food_points
```
viktoria-rybenchuk, line 43: you can write this part in one line. Use `sum` and generator expression

app/main.py
```diff
@@ -1 +1,43 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
```
viktoria-rybenchuk, line 29: Do you use `self` here? Read checklist, for some methods you should use `staticmethod`

viktoria-rybenchuk (changes requested):

pavlejviki (changes requested):

PR #800, opened 2023-12-10
Solution

DanuloM:


app/main.py
```diff
@@ -1 +1,48 @@
-#  write your code here
+class Animal:
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True
+                 ) -> None:
```
Dimosphen1, line 6: Place all arguments on a new line so that the opening and closing parenthesis will not have arguments on the same line.

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Dimosphen1, line 19: There is no need for `else`, the code will not execute after the return statement anyway. Remove `else` from this method.

app/main.py
```diff
@@ -1 +1,48 @@
...
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
```
Dimosphen1, line 24: To improve the readability, place all arguments on the new line in this class and for the `Dog` class as well.

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+
+def feed_animals(animals: list) -> int:
+    count = 0
+    for animal in animals:
+        if animal.is_hungry:
+            animal.feed()
+            count += animal.appetite
+
+    return count
```
Dimosphen1, line 48: Use `sum` and pass there a generator expression to implement this logic in one line. Besides that, there is no need to additionally check if the animal is hungry or not because this check is performed inside of the `feed()` method.

app/main.py
```diff
@@ -1 +1,48 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Dimosphen1, line 41: You could indicate that you expect to obtain a list of `Animal` instances for `animals` in the type annotation.

Dimosphen1 (changes requested):

PR #801, opened 2023-12-14
'Solution' py-feed-animals

yura-push:


app/main.py
```diff
@@ -1 +1,52 @@
...
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 7,
```
Dimosphen1, line 42: You should hardcode appetite in the `super().__init__` and not allow to redefine with argument it in the class `__init__` as mentioned in the task description. Both `Cat` and `Dog` classes should have only two arguments for the `__init__`: `name` and `is_hungry`.

app/main.py
```diff
@@ -1 +1,52 @@
...
+    ) -> None:
+        super().__init__(name, is_hungry=True, appetite=7)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Dimosphen1, line 51: You could indicate that you expect to obtain a list of `Animal` instances for `animals` in the type annotation.

app/main.py
```diff
@@ -1 +1,52 @@
...
+    return sum([animal.feed() for animal in animals if animal.is_hungry])
```
Dimosphen1, line 52: Pass a generator expression instead of a list comprehension to `sum` to improve efficiency. Besides that, there is no need to perform the check for hungriness, it is done in the `feed` method.

Dimosphen1 (changes requested):

pavlejviki (approved):
Good job!

PR #813, opened 2023-12-21
Solution

narizhiy:


app/main.py
```diff
@@ -1 +1,42 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
YukihiroSM, line 18:
```suggestion
        return 0
```
Else is not needed here, cause you have return in if

YukihiroSM (changes requested):
Else looks good! Nicely done!

app/main.py
narizhiy: done

app/main.py
```diff
@@ -1 +1,41 @@
...
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum(
+        [
+            animal.feed() for animal in animals
+        ]
+    )
```
EdAlekseiev, line 41:
```suggestion
    return sum(animal.feed() for animal in animals)
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #815, opened 2023-12-22
Solution

angelzagniy:


app/main.py
```diff
@@ -1 +1,54 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list[Animal]) -> int:
+    sum_of_food = 0
+    for animal in list_of_animals:
+        sum_of_food += animal.feed()
+    return sum_of_food
```
pavlejviki, line 54: simplify this method using sum() and a generator expression

pavlejviki (changes requested):

PR #840, opened 2024-01-13
Solution

MaxymChyncha:


app/main.py
```diff
@@ -1 +1,54 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+
+        return 0
```
NEkropo1, line 22: You could make this with one return, if you create food_points parameter, which will be 0 in 1 case and self_appetite in other

app/main.py
```diff
@@ -1 +1,54 @@
...
+    ) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal | Cat | Dog]) -> int:
+    return sum([animal.feed() for animal in animals])
```
NEkropo1, line 54: you can make this with generator, not list comprehension.

NEkropo1 (changes requested):

PR #842, opened 2024-01-14
Solution

isikach:


app/main.py
```diff
@@ -1 +1,52 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    total_food_points = 0
+    for animal in list_of_animals:
+        if animal.is_hungry:
+            total_food_points += animal.appetite
+            animal.feed()
+    return total_food_points
```
MaxymChyncha, line 52: You can replace this part of code using function sum() with some generator or list comprehension

MaxymChyncha (approved):
Good job!

app/main.py
isikach: Good idea, I change it

app/main.py
```diff
@@ -13,15 +13,16 @@ def print_name(self) -> None:
 
     def feed(self) -> int:
         if self.is_hungry is False:
```
NEkropo1, line 15:
```suggestion
if not self.is_hungry
```

PR #845, opened 2024-01-14
solution

PythonZem:


app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        else:
```
yulias97, line 20: `else` is redundant here

app/main.py
```diff
@@ -1 +1,53 @@
...
+            return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
```
yulias97, line 28: appetite parameter shouldn't be here

app/main.py
```diff
@@ -1 +1,53 @@
...
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
```
yulias97, line 31: `3` should be instead of `appetite` (same for Dog)

yulias97 (changes requested):

NEkropo1 (approved):
Good job!

PR #848, opened 2024-01-15
Solution

oksankakot:


app/main.py
```diff
@@ -1 +1,52 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    food_points = 0
+    for animal in animals:
+        food_points += animal.feed()
+
+    return food_points
```
PythonZem, line 52: Here it would be better to use the sum() function together with compression

PythonZem (commented):
gj!

PythonZem (approved):
gj!

PR #849, opened 2024-01-15
Solution feed animals

mdubyna:


app/main.py
```diff
@@ -1 +1,44 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.appetite = 3
```
PythonZem, line 24: it is rudement code. You can write this value like defolt argument in __init__

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.appetite = 7
```
PythonZem, line 35: it is rudement code. You can write this value like defolt argument in init

PythonZem (approved):
Totally is nice, but have a moment with default attributes

PR #852, opened 2024-01-16
Solution

Bombardier256:
To tell the truth it is not clear how to inherit correct way __init__ method for Cat and Dog  by means of super() for this task.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -39,10 +39,5 @@ class Dog(Animal):
         print("The slippers delivered!")
 
 
-def feed_animals(animals: list) -> int:
-    consumed_food = 0
-
-    for animal in animals:
-        consumed_food += animal.feed()
-
-    return consumed_food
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,43 @@
...
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
vasylhnatiuk, line 43:
```suggestion
    return sum(animal.feed() for animal in animals)
```

vasylhnatiuk: you can avoid it in sum()

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #853, opened 2024-01-16
Solution

Bulleter06:


app/main.py
```diff
@@ -1 +1,39 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> None:
+    total_food_points = 0
+    for animal in animals:
+        total_food_points += animal.feed()
+    return total_food_points
```
oksankakot, line 39: Here will be better to use sum() with compression

NEkropo1, line 39: what about making it 1-liner with some built-in methods?)

NEkropo1 (changes requested):

app/main.py
```diff
@@ -33,7 +33,4 @@ def bring_slippers(self) -> None:
 
 
 def feed_animals(animals: list) -> None:
```
NEkropo1, line 35: I guess it should not return None )

NEkropo1 (changes requested):
LGTM!

NEkropo1 (approved):

PR #856, opened 2024-01-16
'Solution'

NataliiaKorzh:


Anna728560 (approved):

app/main.py
```diff
@@ -1 +1,46 @@
...
+    def print_name(self) -> None:
+        if self.is_hungry is True:
+            print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
NEkropo1, line 18: you dont need this else, also you can try to rewtire it with single return

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_food_points = 0
+
+    for animal in animals:
+        if animal.is_hungry:
+            animal.feed()
+            total_food_points += animal.appetite
+
+    return total_food_points
```
NEkropo1, line 46: best case for 1 liner )

NEkropo1 (changes requested):

PR #859, opened 2024-01-16
"Solution"

achernyshovv:


app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
tskopenko, line 19: delete `else` in ur solution

tskopenko (changes requested):
U can do better!

PR #863, opened 2024-01-16
Solution

tskopenko:


AntonSkyrda (approved):
Looks good, but why method feed returns Any type?

tskopenko:
> Looks good, but why method feed returns Any type?

honestly, it was part of the bugfix.

NEkropo1 (approved):
LGTM!

PR #864, opened 2024-01-16
Task solution

shaiduchyk:


artemlitvenko (approved):

app/main.py
```diff
@@ -1 +1,46 @@
...
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> Any:
```
NEkropo1, line 17: You sure you need Any here?)

app/main.py
```diff
@@ -1 +1,46 @@
...
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
```
NEkropo1, line 22: better case would be single return in a method

NEkropo1 (changes requested):

app/main.py
```diff
@@ -14,12 +14,12 @@ def __init__(self,
...
-    def feed(self) -> Any:
+    def feed(self) -> int:
+        result = self.appetite if self.is_hungry else 0
         if self.is_hungry:
-            print(f"Eating {self.appetite} food points...")
+            print(f"Eating {result} food points...")
             self.is_hungry = False
-            return self.appetite
-        return 0
+        return result
```
NEkropo1, line 22: Great!

NEkropo1 (approved):
Good job!

PR #872, opened 2024-01-20
Solution

nicksetrakov:


app/main.py
```diff
@@ -1 +1,52 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
```
LiudmylaKulinchenko, line 27: Cat\Dog classes should have two arguments - name, is_hungry

app/main.py
```diff
@@ -1 +1,52 @@
...
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
```
LiudmylaKulinchenko, line 30: You can pass 3 directly in a super method. Same for Dog

YuriiKindrat (approved):
LGTM!

PR #894, opened 2024-02-02
solution

liudmylaperchevska:


app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            self.is_hungry = False
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        else:
```
vasylhnatiuk, line 20: you can avoid else here

app/main.py
```diff
@@ -1 +1,47 @@
...
+        super().__init__(name, is_hungry)
+        self.appetite = 7
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(args: list[Animal | Cat | Dog]) -> int:
```
vasylhnatiuk, line 46:
```suggestion
def feed_animals(args: list[Anima]) -> int:
```

vasylhnatiuk: Cat and Dog are Animal

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,47 @@
...
+    return sum([animal.feed() for animal in args])
```
vasylhnatiuk, line 47: you don't need use list comprehension, you can use generator expression for memory efficiency

vasylhnatiuk (changes requested):

app/main.py
liudmylaperchevska: oh, really

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            self.is_hungry = False
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        return int(0)
```
pavlejviki, line 20: 0 is already an int

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #897, opened 2024-02-03
Solution

vsupruniuk:


app/main.py
```diff
@@ -1 +1,46 @@
...
+            return self.appetite
+
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, is_hungry=is_hungry, appetite=3)
+
+    def catch_mouse(self) -> None:
```
LiudmylaKulinchenko, line 28: U can use staticmethod here and in Dog.brint_slippers method

app/main.py
```diff
@@ -1 +1,46 @@
...
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_points = 0
+
+    for animal in animals:
+        total_points += animal.feed()
+
```

app/main.py
```diff
@@ -1 +1,46 @@
...
+            is_hungry: bool = True
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int | None:
```
LiudmylaKulinchenko, line 15: returned value will not be None in any case. You require appetite as int in Animal creation or return 0, so it's integer anyway

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice! ☀

PR #914, opened 2024-02-10
add task solution

EugeneSherepa:


app/main.py
```diff
@@ -1 +1,42 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals])
```
LiudmylaKulinchenko, line 42: `sum` works with generators, so [] are not needed

PR #916, opened 2024-02-12
Solution

CreepyMo:


app/main.py
```diff
@@ -1 +1,45 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = 0
+    for animal in animals:
+        if animal.is_hungry:
+            total_food_points += animal.appetite
+        animal.feed()
+    return total_food_points
```
Dimosphen1, line 45: Use `sum` and pass list comprehension or generator expression there with required operations instead of multiline code. The method should consist of one line and no check whether the animal is hungry or not because it is done in the `feed` method.

app/main.py
```diff
@@ -1 +1,45 @@
...
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Dimosphen1, line 39: Indicate that you expect to obtain a list of `Animal` class instances for `animals`.

app/main.py
```diff
@@ -1 +1,45 @@
...
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
```
Dimosphen1, line 35: You could make this method and `catch_mouse` static since they do not require `self` argument.

app/main.py
```diff
@@ -1 +1,45 @@
-#  write your code here
+class Animal:
+
+    def __init__(self, name: str, appetite: int, is_hungry: bool = True)\
```
Dimosphen1, line 3: Place each argument in the `__init__` on a new line to improve the readability. There should be no arguments on the same line with opening and closing parenthesis. Follow this practice of writing __init__ in other classes as well as avoiding line breaking.

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Dimosphen1, line 17: There is no need for `else` since code will not be executed after the `return` statement anyway. Remove redundant `else`.

Dimosphen1 (changes requested):

app/main.py
CreepyMo: Done

app/main.py
CreepyMo: Done

app/main.py
CreepyMo: Done

app/main.py
CreepyMo: Done

app/main.py
CreepyMo: Done

app/main.py
```diff
@@ -1 +1,53 @@
...
+    ) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
pavlejviki, line 53: no need to use list comprehension, use a generator expression instead

pavlejviki (changes requested):

app/main.py
CreepyMo: Done

Polyakiv-Andrey (approved):
Good job!

PR #924, opened 2024-02-14
Animal classes was created

itpsychologist:


app/main.py
```diff
@@ -1 +1,54 @@
 #  write your code here
+class Animal:
+
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
```
Dimosphen1, line 7: Place each argument on a new line including `self`, so that there would be no arguments on the same lines with opening and closing parenthesis.

app/main.py
```diff
@@ -1 +1,54 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+
+    def __init__(self,
```
Dimosphen1, line 26: Place each argument on a new line including `self`, so that there would be no arguments on the same lines with opening and closing parenthesis.

app/main.py
```diff
@@ -1 +1,54 @@
...
+        self.is_hungry = is_hungry
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self,
```
Dimosphen1, line 39: Place each argument on a new line including `self`, so that there would be no arguments on the same lines with opening and closing parenthesis.

app/main.py
```diff
@@ -1 +1,54 @@
...
+
+
+class Cat(Animal):
+
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.is_hungry = is_hungry
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
```
Dimosphen1, line 31: No need to define arguments before `__init__`, just pass them so the initialization would happen in a parent class.

app/main.py
```diff
@@ -1 +1,54 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Dimosphen1, line 20: There is no need for `else` since the code will not execute after the `return` statement anyway. Remove it.

app/main.py
```diff
@@ -1 +1,54 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    animal_feed = 0
+    for animal in animals:
+        animal_feed += animal.feed()
+    return animal_feed
```
Dimosphen1, line 54: Use `sum` here to perform the same operation in one line along with the return statement. You could pass a list comprehension or generator expression (preferably) to `sum`.

app/main.py
```diff
@@ -1 +1,54 @@
...
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.is_hungry = is_hungry
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Dimosphen1, line 50: Indicate that you expect to obtain a list of `Animal` instances for `animals`.

app/main.py
```diff
@@ -1 +1,54 @@
 #  write your code here
```
Dimosphen1, line 1: Remove this comment

Dimosphen1 (changes requested):

app/main.py
itpsychologist: But then how to return "0"?

vasylhnatiuk (approved):

PR #925, opened 2024-02-15
Solution

olga-kim7:


app/main.py
```diff
@@ -1 +1,49 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+
```
Dimosphen1, line 24: There is no need for `else`, the code will not execute after the `return` statement anyway. Remove redundant `else`

app/main.py
```diff
@@ -1 +1,49 @@
...
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    def catch_mouse(self) -> None:
```
Dimosphen1, line 32: You could use `staticmethod` decorator for methods that do not utilize `self`. Apply it, please

app/main.py
```diff
@@ -1 +1,49 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
```
Dimosphen1, line 41: You could use `staticmethod` decorator for methods that do not utilize `self`. Apply it, please

app/main.py
```diff
@@ -1 +1,49 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: List[Animal]) -> int:
+    total_food_points = 0
+    for animal in animals:
+        total_food_points += animal.feed()
+    return total_food_points
```
Dimosphen1, line 49: Implement this method in one line using `sum`. You could pass there a list comprehension or a generator expression (better).

Dimosphen1 (changes requested):

olga-kim7:
виправлено код) дякую

PR #930, opened 2024-02-19
feed_Solution

OleksandrHubar:


app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Dimosphen1, line 21: There is no need for `else`, code will not execute after `return` anyway. Remove `else`

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_food_point = 0
+
+    for animal in animals:
+        if animal.is_hungry:
+            food_point = animal.feed()
+            total_food_point += food_point
+    return total_food_point
```
Dimosphen1, line 48: Implement this function in one line using `sum`. Besides that, there is no need to perform the check for hungriness, it is done in the `feed` method.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
```
Dimosphen1, line 36: You could add a `staticmethod` decorator to functions that do not utilize the `self`. Add it in the same cases in the future

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #931, opened 2024-02-20
Solution

AlexTsikhun:
I thought that in super().__init()__ we should pass all parameters of father class (in case, if child class has same args and couple new one, like this https://github.com/AlexTsikhun/py-elves-and-dwarves/blob/develop/app/players/dwarves/dwarf.py)
In child class I have 2 params, in super should has at least same number of params? it's related?
`    def __init__(self, name: str, is_hungry: bool = True) -> None:
        super().__init__(name, appetite=3)`

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,48 @@
-#  write your code here
+
+class Animal:
+    def __init__(
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
+        self.name = name
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+        self.name = name
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum(animal.feed() for animal in animals)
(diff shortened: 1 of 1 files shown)


Dimosphen1:
The number of arguments that are passed to the child class may differ from the one present in the parent class. The main point is that `super().__init__()` call should satisfy all requirements to initialize the parent class

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Dimosphen1, line 21: There is no need for else, the code will not execute after `return` anyway. Remove `else`

app/main.py
```diff
@@ -1 +1,48 @@
...
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3)
```
Dimosphen1, line 27: This logic is not right, in case you take `is_hungry` argument, you should pass it to the `super().__init__(` as well. There is no need to redefine arguments on lines 28 and 29/ Make similar changes in the `Dog` class as well

Dimosphen1 (changes requested):

AlexTsikhun:
> The number of arguments that are passed to the child class may differ from the one present in the parent class. The main point is that `super().__init__()` call should satisfy all requirements to initialize the parent class

that is, so that the child class has all the required args from parent (and in the same order)?

app/main.py
AlexTsikhun: > This logic is not right, in case you take `is_hungry` argument, you should pass it to the `super().__init__(` as well. 

I should pass all argument, that take my __init__ in super().__init__?

AlexTsikhun: >There is no need to redefine arguments on lines 28 and 29/ Make similar changes in the `Dog` class as well

`    def __init__(self, name: str, is_hungry: bool = True) -> None:
        super().__init__(name, appetite=3, is_hungry=True)
        self.name = name
        self.is_hungry = is_hungry

    @staticmethod
    def catch_mouse() -> None:
        print("The hunt began!")


class Dog(Animal):
        super().__init__(name, appetite=7, is_hungry=True)
        self.name = name
        self.is_hungry = is_hungry`
  
if I remove `self.is_hungry = is_hungry` from Dog - all right!
but i I remove `self.is_hungry = is_hungry` from Cat - error (    def test_feed_animals_function(animals, total_food_points):
>       assert main.feed_animals(animals) == total_food_points
E       assert 3 == 0
E         +3
E         -0
)

viktoria-rybenchuk (changes requested):

vsmutok (approved):
good job!

PR #942, opened 2024-02-25
Task completed ✅

Githi54:
🐶🐱Implement parent class Animal
🐶 Implement the Dog class
🐱 Implement the Cat class

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,52 @@
-#  write your code here
+from typing import List
+
+
+class Animal:
+    def __init__(
+        self,
+        name: str,
+        appetite: int,
+        is_hungry: bool = True
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,52 @@
...
+
+def feed_animals(animals: List[Animal]) -> int:
+    food_points = 0
+
+    for animal in animals:
+        if animal.is_hungry:
+            food_points += animal.appetite
+            animal.feed()
+
+    return food_points
```
LiudmylaKulinchenko, line 52: Try to rewrite in one line, use `sum`

app/main.py
```diff
@@ -1 +1,49 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: List[Animal]) -> int:
+    return sum(
+        animal.feed()
+        for animal in animals
+        if animal.is_hungry
```
LiudmylaKulinchenko, line 48: If animal is not hungry, animal.feed method will return 0, so not necessary to add this condition here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🌷

PR #948, opened 2024-02-26
Solution

Viesich:


app/main.py
```diff
@@ -1 +1,45 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list) -> int:
```
Dimosphen1, line 41: Do not use type in the variable name in case it is not necessary. Replace `animal_list` with `animals` and indicate that you expect to obtain a list of `Animal` class instances for `animals` in the type annotation

app/main.py
```diff
@@ -1 +1,45 @@
...
+    total_points = 0
+    for animal in animal_list:
+        total_points += animal.feed()
+    return total_points
```
Dimosphen1, line 45: Implement the body of this function in one line using `sum`

Dimosphen1 (changes requested):

PR #951, opened 2024-02-27
Solution

Sergunshot:


app/main.py
```diff
@@ -1 +1,45 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
EdAlekseiev, line 20:
```suggestion
        if self.is_hungry is True:
            print(f"Eating {self.appetite} food points...")
            self.is_hungry = False
            return self.appetite
        return 0
```

app/main.py
```diff
@@ -1 +1,45 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: List[Animal]) -> int:
+    total_point = 0
+    for animal in animals:
+        total_point += animal.feed()
+    return total_point
```
EdAlekseiev, line 45: Rewrite this function to one line with using function `sum`

EdAlekseiev (changes requested):

app/main.py
Sergunshot: Done

app/main.py
Sergunshot: Done

EdAlekseiev (approved):

PR #960, opened 2024-03-05
'Solution'

Vitalii-Haponiuk:


app/main.py
```diff
@@ -1 +1,42 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Dog | Cat | Animal]) -> int:
```
Dimosphen1, line 41: Indicate that you expect to obtain a list of `Animal` class instances for `animals` in the type annotation. `Dog` and `Cat` would be eligible since they inherit from `Animal`

app/main.py
```diff
@@ -1 +1,42 @@
...
+    return sum(beast.feed() for beast in animals if beast.is_hungry)
```
Dimosphen1, line 42: There is no need to perform if check in the generator expression since it is done in the `feed` method. Remove this check

app/main.py
```diff
@@ -1 +1,42 @@
...
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
```
Dimosphen1, line 16: You could simplify your check
```suggestion
        if self.is_hungry:
```

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #968, opened 2024-03-10
Solution

Bujhm99:


app/main.py
```diff
@@ -1 +1,38 @@
...
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.appetite == 0 or not self.is_hungry:
```
LiudmylaKulinchenko, line 12: `if not self.is_hungry` could probably be enough here. Consider to remove a first part of the condition

app/main.py
```diff
@@ -1 +1,38 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(zoo: list) -> int:
+    return sum([animsl.feed() for animsl in zoo])
```
LiudmylaKulinchenko, line 38: `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
GJ!

PR #971, opened 2024-03-13
Solution

Vova-Stoliar:


app/main.py
```diff
@@ -1 +1,52 @@
...
+            return 0
+
+        food_points_to_eat = self.appetite
+
+        print(f"Eating {food_points_to_eat} food points...")
+
+        self.is_hungry = False
+        self.appetite -= food_points_to_eat
+
+        return food_points_to_eat
```
LiudmylaKulinchenko, line 27: `feed - should print Eating {appetite} food points..., set is_hungry to False and return number of eaten food points if animal is hungry. Otherwise, it should return 0 and print nothing.` -- your function has slightly different logic

LiudmylaKulinchenko (changes requested):

app/main.py
Vova-Stoliar: Sorry, do not understand why the function has a different logic
It seems right to me

Vova-Stoliar: Sorry, do not understand
It seems right to me

LiudmylaKulinchenko: I meant, you don't need to change appetite and is_hungry for current animal, according to the task description. But if you think it's appropriate to add here, it's fine, then I have a different comment 😁 Try to avoid creating additional `food_points_to_eat` variable here

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,48 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    food_points = 0
+
+    for animal in animals:
+        food_points += animal.feed()
+
+    return food_points
```
LiudmylaKulinchenko, line 48: Try to rewrite in one line

LiudmylaKulinchenko (changes requested):

PR #974, opened 2024-03-14
Solution!

scandriy:


app/main.py
```diff
@@ -1 +1,50 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    food_counter = 0
+    for beast in animals:
+        if isinstance(beast, Animal):
+            food_counter += beast.feed()
+    return food_counter
```
EdAlekseiev, line 50: Rewrite this function to one line with using function `sum`

EdAlekseiev (changes requested):

app/main.py
scandriy: sure, no problem

EdAlekseiev (approved):
Good job

PR #986, opened 2024-03-20
Develop

Gilgumish:


app/main.py
```diff
@@ -1 +1,41 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
panicua, line 21: you don't need ```else``` just ```return 0``` in this condition cause you'll have ```return 0```, anyway

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
```
panicua, line 29: this method only prints static message so better use @staticmethod I think.

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
```
panicua, line 37: same here, static print

panicua (changes requested):

panicua (approved):
gj

Oleksl888 (approved):
good job!

PR #990, opened 2024-03-21
Solution

y-kondrashova:


app/main.py
```diff
@@ -1 +1,51 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    all_feed_points = 0
+    for animal in animals:
+        if animal.is_hungry:
+            animal.feed()
+            all_feed_points += animal.appetite
+    return all_feed_points
```
CharmNiovi, line 51: You can do it in one line.

app/main.py
```diff
@@ -1 +1,51 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True,
+                 appetite: int = 3) -> None:
```
CharmNiovi, line 25: Write Cat class which is a child of Animal. It should have the __init__ method with two arguments:

-  name - the name of a cat
-  is_hungry - with True default value
There you have three arguments.

app/main.py
```diff
@@ -1 +1,51 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True,
+                 appetite: int = 7) -> None:
```
CharmNiovi, line 37: The last class you should implement is a Dog class. Its __init__ method should have two arguments:

- name - the name of a dog
- is_hungry - with True default value

The same problem

CharmNiovi (changes requested):

app/main.py
```diff
@@ -1 +1,44 @@
...
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([animal.feed() for animal in animals if animal.is_hungry])
```
Oleksl888, line 44: You don't need to check if the animal is hungry here - this will be handled by feed method

Oleksl888, line 44: You don't need to check if the animal is hungry here - this will be handled by feed method

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ!

PR #998, opened 2024-03-22
'Solution1'

yossosh:


alonasorochynska (approved):
good job!

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
YuriiKindrat, line 18: why do u need else here ?

YuriiKindrat (changes requested):

app/main.py
yossosh: Бо у нас в умові задачі вказано. 

Animal should have two methods:

print_name - should print name in the following format: Hello, I'm {name}
feed - should print Eating {appetite} food points..., set is_hungry to False and return number of eaten food points if animal is hungry. _Otherwise, it should return 0 and print nothing._

YuriiKindrat: u could remove else condition and write return 0 on the same level as if condition 
`if <some condition>:`
`        return smth`
`return smth`
so this will work fine without else

app/main.py
```diff
@@ -1 +1,45 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = 0
+    for animal in animals:
+        food_points = animal.feed()
+        total_food_points += food_points
+    return total_food_points
```
YuriiKindrat, line 45: this could be done in one line, try to use sum method, with generator

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ!

PR #1001, opened 2024-03-23
Solution#1

VitalyBashkiser:


siaserb (approved):
Well done! But in the last function you can use sum().

YuriiKindrat (approved):
LGTM!

PR #1005, opened 2024-03-24
Solution

onyevyerov:


app/main.py
```diff
@@ -1 +1,51 @@
...
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list = None) -> str or int:
+    if animals is None:
```
pavlopro, line 46: Please simplify this, as it's understood that this method takes a list of animals.
Also do not forget to change annotations

pavlopro (changes requested):

PR #1007, opened 2024-03-25
'Solution'

afaidiuk:


VladyslavKsh (approved):
Good job

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Oleksl888, line 18: No need for else here

app/main.py
```diff
@@ -1 +1,44 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Oleksl888, line 40: be more specific about types of list elements

app/main.py
```diff
@@ -1 +1,44 @@
...
+    total_food_points = 0
+    for animal in animals:
+        total_food_points += animal.feed()
+    return total_food_points
```
Oleksl888, line 44: you can calculate total in one line here using  generator expression and `sum` function

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,41 @@
...
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_food_points = sum(animal.feed() for animal in animals)
+    return total_food_points
```
YuriiKindrat, line 41: why do u need to create variable here, could u just return sum(...)

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🔥

PR #1008, opened 2024-03-25
Solution

Ruxxell54:


app/main.py
```diff
@@ -1 +1,40 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
YuriiKindrat, line 16: why do u need else here ?

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = 0
+    for animal in animals:
+        total_food_points += animal.feed()
+    return total_food_points
```
YuriiKindrat, line 40: this could be done in one line, using sum method

YuriiKindrat (changes requested):

PR #1009, opened 2024-03-25
'Solution'

dkotkod:


app/main.py
```diff
@@ -1 +1,44 @@
...
+        super().__init__(name, 3, is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.name = name
+        self.is_hungry = is_hungry
```
YuriiKindrat, line 33: why do u need this ?

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> None:
+    total = 0
+    for animal in animals:
+        total += animal.feed()
+    return total
```
YuriiKindrat, line 44: try to do this in one line using sum method

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,39 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    def catch_mouse(self) -> None:
```
Oleksl888, line 26: How should we define a method that doesn't require `self` as a parameter? Check for `bring_slippers` as well.

app/main.py
```diff
@@ -1 +1,39 @@
...
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Oleksl888, line 38: what about types for list elements?

Oleksl888 (changes requested):

app/main.py
Oleksl888: I think you need to focus - still no `staticmethod` here

Oleksl888 (changes requested):

Oleksl888 (approved):
you made it!

PR #1034, opened 2024-04-08
Solution

mlupekha:


app/main.py
```diff
@@ -1 +1,42 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
EdAlekseiev, line 18:
```suggestion
        if self.is_hungry:
            print(f"Eating {self.appetite} food points...")
            self.is_hungry = False
            return self.appetite
        return 0
```

app/main.py
```diff
@@ -1 +1,42 @@
...
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_needed = 0
+    for animal in animals:
+        if animal.is_hungry:
+            total_food_needed += animal.feed()
+    return total_food_needed
```
EdAlekseiev, line 42: Rewrite this function with using function `sum`

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1035, opened 2024-04-10
Solution

irinastetsko:


app/main.py
```diff
@@ -1 +1,45 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> None:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
EdAlekseiev, line 20:
```suggestion
        if self.is_hungry:
            print(f"Eating {self.appetite} food points...")
            self.is_hungry = False
            return self.appetite
        return 0
```

app/main.py
```diff
@@ -1 +1,45 @@
...
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = 0
+    for animal in animals:
+        if animal.is_hungry:
+            total_food_points += animal.feed()
+    return total_food_points
```
EdAlekseiev, line 44: Rewrite this method to one line

EdAlekseiev (changes requested):

app/main.py
EdAlekseiev: Still not fixed

EdAlekseiev (changes requested):

PR #1040, opened 2024-04-15
Done

Stekloduv:


app/main.py
```diff
@@ -1 +1,44 @@
-#  write your code here
+class Animal:
+    def __init__(self
+                 , name: str
+                 , appetite: int
+                 , is_hungry: bool = True) -> None:
```
Dimosphen1, line 5: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability. Do not place commas on the first position, place them on the last position in line instead

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Dimosphen1, line 18: There is no need for `else` since the code will not execute after the `return` statement anyway. Please, remove the redundant `else`

app/main.py
```diff
@@ -1 +1,44 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_ls: list) -> int:
+    points = 0
+    for animal in animal_ls:
+        points += animal.feed()
+    return points
```
Dimosphen1, line 44: Implement this method in one line using `sum` and passing generator expression there

Dimosphen1 (changes requested):

PR #1044, opened 2024-04-21
Solution

GGsosna:


app/main.py
```diff
@@ -1 +1,52 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
vsmutok, line 24:
```suggestion
            return self.appetite
        return 0
```
no need write `else` in this case

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1051, opened 2024-04-27
triukhan solution

triukhan:


app/main.py
```diff
@@ -1 +1,47 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True,
+                 ) -> None:
```
LiudmylaKulinchenko, line 38: fix formatting in both init methods

LiudmylaKulinchenko (changes requested):

PR #1063, opened 2024-05-20
Solution

SuskyiVolodymyr:


DevSpaciX (approved):
Good work ! But Imho -> this way func looks better ->
```
if self.is_hungry:
            self.is_hungry = False
            return self.appetite
        return 0

```

PR #1064, opened 2024-05-22
solution

AndrewwSSS:


app/main.py
```diff
@@ -1 +1,37 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str,
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
```
DevSpaciX, line 9: Can you make this method static ? I am about `@staticmethod` : ) If you do not using `self` in your function it is better to use decorator  `@staticmethod`

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1 +1,37 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    def catch_mouse(self) -> None:
```
DevSpaciX, line 24: Can you make this method static ? I am about `@staticmethod` : ) If you do not using `self` in your function it is better to use decorator  `@staticmethod`

AndrewwSSS: done

DevSpaciX (approved):

PR #1065, opened 2024-05-22
Solution

necobra:


app/main.py
```diff
@@ -1 +1,40 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            feeded = self.appetite
+            self.appetite = 0
```
DevSpaciX, line 16: image
So you do not need this code on line 15 and 16 . Or why did you add it ?)

necobra: didn't fully understand the logic in the first time)

DevSpaciX (approved):

PR #1066, opened 2024-05-22
py-feed-animals v1

BornToLivee:


BornToLivee:
sometimes PyCharm tells me, that some methods can be "@staticmethod". Should ignore this or add "@staticmethod" is a good practice?

DevSpaciX:
> sometimes PyCharm tells me, that some methods can be "@staticmethod". Should ignore this or add "@staticmethod" is a good practice?

Yeah , if method do not using `self` statement , it is better to make the method static

DevSpaciX (changes requested):
Ss make your methods static using @staticmethod where it needed : )

DevSpaciX (approved):

PR #1069, opened 2024-05-24
solution

VShveda:


app/main.py
```diff
@@ -1 +1,45 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
Oleg-s-s, line 41: list of what

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1 +1,45 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    food_point = 0
```
DevSpaciX, line 42: You can try to do this logic in 1 line using `sum()` method and `comprehension`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1073, opened 2024-05-26
add solution

jpdreamthug:


app/main.py
```diff
@@ -1 +1,47 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
```
Oleg-s-s, line 18: I mean... its ok, but what if your logic was 100 lines of code?
In such scenarios you can think in reverse, f.e.
`if not self.is_hungry` 
and after main logic

app/main.py
```diff
@@ -1 +1,47 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    hunger_points = 0
+    for animal in animals:
+        if animal.is_hungry:
+            animal.feed()
+            hunger_points += animal.appetite
+    return hunger_points
```
Oleg-s-s, line 47: use `sum()` and avoid __standart__ for loop

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1 +1,42 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum(animal.feed() if animal.is_hungry else 0 for animal in animals)
```
DevSpaciX, line 42: You already put 0 in the appetite viriable `if animal.is_hungry` on line 15 . Is not you ?) So you repeating yourself , pls fix that ))

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1075, opened 2024-05-26
Solution

dmytrostepanenko:


app/main.py
```diff
@@ -1 +1,54 @@
...
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(
+                "Eating {appetite} food points..."
+                .format(appetite=self.appetite)
+            )
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
vsmutok, line 22:
```suggestion
        if not self.is_hungry:
            return 0

        print(f"Eating {self.appetite} food points...")
        self.is_hungry = False
        return self.appetite
```

app/main.py
```diff
@@ -1 +1,54 @@
...
+        self.appetite = 7
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
vsmutok, line 54:
```suggestion
    return sum(animal.feed() for animal in animals)
```
no need to create list in this case

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1077, opened 2024-05-26
Solution

SrMusienko:


app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
DevSpaciX, line 20: You don’t need `else` if you have a `return` statement in `if`.

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1078, opened 2024-05-26
'Solution'

BohdanKuzik:


app/main.py
```diff
@@ -1 +1,44 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_points = 0
```
DevSpaciX, line 41: You can try to use `sum()` function with the `comprehension` ( It will be 1 line func )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1080, opened 2024-05-27
Solution

Andrii1000:


app/main.py
```diff
@@ -1 +1,43 @@
-#  write your code here
+from typing import List
+
+
+class Animal:
+    def __init__(self, name: str,
+                 appetite: int,
+                 is_hungry: bool = True
+                 ) -> None:
```
Andrii1000, line 8: think is bad code style. must be:
```suggestion
                def __init__(
                self, name: str,
                 appetite: int,
                 is_hungry: bool = True
                 ) -> None:
```

DevSpaciX: good point

DevSpaciX (approved):

PR #1086, opened 2024-05-27
'Solution'

Oleh323:


app/main.py
```diff
@@ -1 +1,45 @@
...
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if not self.is_hungry:
+            return 0
+        else:
```
DevSpaciX, line 16: You do not need `else` here if you have `return` in `if statement `

app/main.py
```diff
@@ -1 +1,45 @@
...
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    result = 0
+    for char in list_of_animals:
```
DevSpaciX, line 42: You can try to do this logic in 1 line using `sum()` method and `comprehension`

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1087, opened 2024-05-27
'Solution'

Irina17191:


app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
```
Oleg-s-s, line 24: its ok but in cases like this it will be more understandable to write reversed
`if not self.is_hungry`

app/main.py
```diff
@@ -1 +1,50 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_appetite = 0
+    for animal in animals:
+        total_appetite += animal.feed()
+
+    return total_appetite
```
Oleg-s-s, line 50: Use `sum()` and write it in 1 line

Oleg-s-s (changes requested):

app/main.py
Irina17191: Thank you, changed it

Oleg-s-s (approved):
👍

PR #1088, opened 2024-05-27
sol1

Unlie9:


app/main.py
```diff
@@ -1 +1,57 @@
...
+
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True
+                 ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+        if is_hungry:
```
DevSpaciX, line 12: You do not need to store hungry animals in separeted list . You just can use the `is_hungry` attribute

app/main.py
```diff
@@ -1 +1,57 @@
-#  write your code here
+class Animal:
+    hungry_animals = []
```
DevSpaciX, line 2: You do not need this : )

app/main.py
```diff
@@ -1 +1,57 @@
...
+            Animal.hungry_animals.append(self)
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> None | int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            if self in Animal.hungry_animals:
```
DevSpaciX, line 22: Alos this is redundant logic : ) Just `return self.appetite`

app/main.py
```diff
@@ -1 +1,57 @@
...
+                Animal.hungry_animals.remove(self)
+            return self.appetite
+        else:
```
DevSpaciX, line 25: You don’t need `else` if you have a `return` statement in `if.`

app/main.py
```diff
@@ -1 +1,57 @@
...
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(hungry_animals: list) -> None:
+    needed_food = 0
+
+    for hungry_animal in hungry_animals:
```
DevSpaciX, line 54: You can try to use `sum()` function with the comprehension ( It will be 1 line func )

app/main.py
```diff
@@ -1 +1,57 @@
...
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+        if is_hungry:
+            Animal.hungry_animals.append(self)
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> None | int:
```
DevSpaciX, line 18: `0` it is `int` too , so you can leave just `int` : )

app/main.py
```diff
@@ -1 +1,57 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(hungry_animals: list) -> None:
```
DevSpaciX, line 51: Hm , why you returning `None`? If you have some `calculation` logic ?)

DevSpaciX (changes requested):

DevSpaciX (approved):
Great now )

PR #1089, opened 2024-05-27
Develop

b3v3kt0r:


app/main.py
```diff
@@ -1 +1,50 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
DevSpaciX, line 20: You don’t need `else` if you have a return statement in `if`.

app/main.py
```diff
@@ -1 +1,50 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name)
+        self.is_hungry = is_hungry
+        self.appetite = 7
```
DevSpaciX, line 39: You can provide this params in `__init__` method on 37 line . Is not you ?)

app/main.py
```diff
@@ -1 +1,50 @@
...
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name)
+        self.is_hungry = is_hungry
+        self.appetite = 3
```
DevSpaciX, line 28: Same ) Just provide this params in `__init__ `

app/main.py
```diff
@@ -1 +1,50 @@
...
+class Animal:
+    animals = []
+
+    def __init__(self, name: str,
+                 appetite: int = 0,
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+        Animal.animals.append(self)
```
DevSpaciX, line 10: Actually you dont need to create list with the animals , just remove it )

app/main.py
```diff
@@ -1 +1,50 @@
...
+        self.is_hungry = is_hungry
+        self.appetite = 7
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    hungry = 0
```
DevSpaciX, line 47: You can try to do this logic in 1 line using `sum()` method and `comprehension `

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1092, opened 2024-05-27
'Solution'

VitaliiDats1:


app/main.py
```diff
@@ -1 +1,44 @@
...
+    def catch_mouse():
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers():
```
DevSpaciX, line 36: Fix all `flake8` errors about type hints : )
If you do not have a `return` statement that is mean you returning `nothing`

app/main.py
```diff
@@ -1 +1,44 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    points = 0
```
DevSpaciX, line 41: You can try to do this logic in 1 line using `sum()` method and `comprehension`

DevSpaciX (changes requested):
Add type annotations : ) And addtional fix in the comment

DevSpaciX (approved):
Great !

PR #1094, opened 2024-05-28
Done task!

TkachukTara:


app/main.py
```diff
@@ -1 +1,44 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> None:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            food_point = self.appetite
+            self.is_hungry = False
+            return food_point
+        return 0
```
DevSpaciX, line 18: You do not need extra viriable `food_point` , just return `self.appetite`

app/main.py
```diff
@@ -1 +1,44 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    sum_of_point_food = 0
+    for animal in animals:
+        if animal.is_hungry is True:
+            sum_of_point_food += animal.feed()
+    return sum_of_point_food
```
DevSpaciX, line 44: You can do it in one line using `sum()`

DevSpaciX (changes requested):

DevSpaciX (approved):
Good job!

PR #1098, opened 2024-05-29
'Solution'

hhorhe:


app/main.py
```diff
@@ -1 +1,43 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
EdAlekseiev, line 18: `else` block is redundant, because if contains `return`.
```suggestion
        if self.is_hungry:
            print(f"Eating {self.appetite} food points...")
            self.is_hungry = False
            return self.appetite
        return 0
```

app/main.py
```diff
@@ -1 +1,43 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    accum = 0
+    for animal in animals:
+        if animal.is_hungry:
+            accum += animal.feed()
+
+    return accum
```
EdAlekseiev, line 43: Rewrite this method to one line with using `sum`

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,37 @@
...
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
vasylhnatiuk, line 36:
```suggestion
def feed_animals(animals: list[Animal]) -> int:
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1100, opened 2024-05-30
Solution

Katy4S:


app/main.py
```diff
@@ -1 +1,43 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
vsmutok, line 20:
```suggestion
            return self.appetite
        return 0
```
`else` redundant in this case

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: List[Animal]) -> int:
+    total_food_points: int = 0
+    for animal in animals:
+        total_food_points += animal.feed()
+    return total_food_points
```
vsmutok, line 43: try to write in in one line
use `sum`

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,39 @@
-#  write your code here
+from typing import List
+
+
+class Animal:
+    def __init__(self, name: str, appetite: int,
+                 is_hungry: bool = True) -> None:
```
YukihiroSM, line 6: If you are separating parameters into lines - it's better to have one parameter for one line.

YukihiroSM (approved):
Looks pretty good to me. There is one tiny comment - please, consider it in future :)

PR #1117, opened 2024-06-19
Solution

Alex414414:


app/main.py
```diff
@@ -1 +1,58 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = 0
+    for animal in animals:
+        total_food_points += animal.feed()
+    return total_food_points
```
vsmutok, line 58: try write it in one line
use `sum`

app/main.py
```diff
@@ -1 +1,58 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
vsmutok, line 21:
```suggestion
            return self.appetite
        return 0
```
`else` redundant in this case

vsmutok (changes requested):

EdAlekseiev (approved):

PR #1119, opened 2024-06-20
Add solution

MRozdoba:


app/main.py
```diff
@@ -1 +1,55 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
+            is_hungry: bool = True
+    ) -> None:
```
Dimosphen1, line 29: The `__init__` of `Cat` and `Dog` classes should take only two arguments according to the task description. Remove `appetite` and hardcode it in the `super().__init__()` in both classes

app/main.py
```diff
@@ -1 +1,55 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    sum_food = 0
+    for animal in animals:
+        sum_food += animal.feed()
+    return sum_food
```
Dimosphen1, line 55: You could implement this method in one line using `sum`

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1120, opened 2024-06-21
Solution

Wallian169:


pavlopro (changes requested):

app/main.py
```diff
@@ -1 +1,42 @@
...
+    def __init__(self, name: str, is_hungry: bool = True,) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals if animal.is_hungry])
```
pavlopro, line 42: You can omit [] and is_hungry check as method already has it

pavlopro (approved):
Good Job!

app/main.py
Wallian169: ok

PR #1121, opened 2024-06-21
Develop

HalynaRohatska:


app/main.py
```diff
@@ -1 +1,48 @@
...
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is True:
```
MNDonut, line 19: you can just use "if self.is_hungry" because it's True or False

app/main.py
```diff
@@ -1 +1,48 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = not self.is_hungry
```
MNDonut, line 21: this is technically correct, but in this case you definitely need to set only False, so it'll be more explicit for other devs.
"not self.is_hungry" is often used when it's possible to switch between True and False anytime. For example when you like a photo in instagram. There is a field "is_liked". Each time user clicks on a heart, it should set vice-versa value. Like -> unlike, unlike -> Like. In your case only the first case is possible ;)

app/main.py
```diff
@@ -1 +1,48 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
```
MNDonut, line 32: Why is it static method? In the example we call this method from class instance "cat2.catch_mouse()"

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
```
MNDonut, line 41: Why is it static method? In the example we call this method from class instance

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_points_of_food = 0
+    for animal in animals:
+        total_points_of_food += animal.feed()
+    return total_points_of_food
```
MNDonut, line 48: Just for your information, you can use here sum() function and list comprehension to do it in one line, but that's okay :)

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_points_of_food = 0
+    for animal in animals:
+        total_points_of_food += animal.feed()
+    return total_points_of_food
```
LiudmylaKulinchenko, line 46: Let's still try to do this in one line! Note that `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

MNDonut (approved):
Super clear code!

PR #1128, opened 2024-06-28
Solution

r0manvynnytskyi:


app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
vasylhnatiuk, line 18: else is redundant here

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list[Animal]) -> int:
```
vasylhnatiuk, line 40:
```suggestion
def feed_animals(animals: list[Animal]) -> int:
```

app/main.py
```diff
@@ -1 +1,44 @@
...
+    total_food_points = 0
+    for animal in list_of_animals:
```
vasylhnatiuk, line 42: use list comprehension

vasylhnatiuk (changes requested):

app/main.py
r0manvynnytskyi: > use list comprehension

got it, I just pushed the differences, but I can`t request the changes to you

vasylhnatiuk (approved):

PR #1129, opened 2024-06-28
implemented Animal OOP realisation with "unique" constructors

MasakDirt:


app/main.py
```diff
@@ -1 +1,37 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str, appetite: int,
+                 is_hungry: bool = True) -> None:
```
MNDonut, line 3: It's a bad idea to add parameters in this way. Either all of them should be in 1 line or 1 parameter - 1 line

MNDonut (changes requested):
Some PEP8 improvements :)

oliasymak24 (approved):

PR #1136, opened 2024-07-05
Solution Animal Feed

Alex-Plugin:


app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    food_points = sum(
+        [animal.feed() for animal in animals]
+    )
+    return food_points
```
vsmutok, line 48:
```suggestion
    return sum(animal.feed() for animal in animals)
```

app/main.py
```diff
@@ -1 +1,48 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        else:
+            return 0
```
vsmutok, line 23:
```suggestion
            return self.appetite
        return 0
```
`else` redundant in this case

vsmutok (changes requested):

app/main.py
Alex-Plugin: Thank you for the correction! Very elegant solution!

app/main.py
Alex-Plugin: Thank you!

PR #1142, opened 2024-07-11
Implemented solution (added all classes and did refactor in order not to copy the code when I already had the function with same logic)

dimak20:


app/main.py
```diff
@@ -1 +1,40 @@
-#  write your code here
+class Animal:
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
```
MNDonut, line 5: We recommend to either add all parameters in 1 line or 1 parameter - 1 line

app/main.py
```diff
@@ -1 +1,40 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
```
MNDonut, line 27: Why static method? Is there a requirement for it?

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
```
MNDonut, line 36: Why static method? Is there a requirement for it?

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+
+def feed_animals(animals: list) -> int:
+    return sum([Animal.feed(animal) for animal in animals])
```
MNDonut, line 40: You don't need to call feed method in this way. As Dog and Cat inherit from Animal, they both have the same attributes/methods that the parent class has, so you can use animal's method to feed the animal :)

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,40 @@
...
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
```
oliasymak24, line 25: Will be better to use keywords arguments.

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
```
oliasymak24, line 33: Will be better to use keywords arguments.

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
oliasymak24, line 39: Pay attention to type hints.

oliasymak24 (approved):
You changed the previous comments well, pay extra attention to the passing arguments and type hints :)

PR #1146, opened 2024-07-14
Solution

OleksiukStepan:
Created and implemented :

- class **Animal**
- class **Cat**
- class **Dog**
- method **feed_animals**

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,42 @@
-#  write your code here
+class Animal:
+    def __init__(
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,42 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
```
MNDonut, line 29: Why static method? Is there a requirement to make it static?

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
```
MNDonut, line 38: Why static method? Is there a requirement to make it static?

MNDonut (changes requested):

oliasymak24 (approved):
Good job!

PR #1148, opened 2024-07-19
Solution

DSahalatyi:


app/main.py
```diff
@@ -1 +1,56 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_appetite = 0
+    for animal in animals:
+        if animal.is_hungry:
+            total_appetite += animal.feed()
+    return total_appetite
```
LiudmylaKulinchenko, line 56: You can use `sum` function here

LiudmylaKulinchenko (changes requested):

MNDonut (approved):

PR #1155, opened 2024-07-26
'Solution'

dartomOOv:


app/main.py
```diff
@@ -1 +1,44 @@
...
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = not self.is_hungry
```
MNDonut, line 19: here you can set explicity False instead of using `not`. we use `not` when we should set vice-versa values, but in this case we shouldn't because of `if self.is_hungry condition`

MNDonut (changes requested):

PR #1156, opened 2024-07-26
implement py-feed-animals

ValeriiaKyr:


app/main.py
```diff
@@ -1 +1,60 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is False:
+            return 0
+        else:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
```
LiudmylaKulinchenko, line 21: No need to use else if you have return in your if statement

app/main.py
```diff
@@ -1 +1,60 @@
...
+
+def feed_animals(list_feed: list) -> int:
+    needed = 0
+    for animal in list_feed:
+        if animal.is_hungry:
+
+            needed += animal.appetite
+            Animal.feed(animal)
+
+    return needed
```
LiudmylaKulinchenko, line 60: U can rewrite it in one line using sum

app/main.py
```diff
@@ -1 +1,60 @@
...
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_feed: list) -> int:
```
LiudmylaKulinchenko, line 52: Let's also annotate what type of objects we expect to have in list

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌸

PR #1158, opened 2024-07-26
implement Animal class and its subclasses

vladislav-tsybuliak1:


app/main.py
```diff
@@ -1 +1,42 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
LiudmylaKulinchenko, line 42: No need to use [] since sum works with generators

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌸

PR #1160, opened 2024-07-27
Task Solution

achalaidiuk:


app/main.py
```diff
@@ -1 +1,50 @@
...
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        result = self.appetite if self.is_hungry else 0
+        if self.is_hungry:
```
LiudmylaKulinchenko, line 17: To avoid duplicating conditions and creating a new value, you can remove ternary operator and put this logic inside of the `if self.is_hungry` condition

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            result = self.appetite
+            print(f"Eating {result} food points...")
+            self.is_hungry = False
+        else:
+            result = 0
+        return result
+
```
LiudmylaKulinchenko, line 23: How about just return value insine of `if` to remove a few lines?

LiudmylaKulinchenko (changes requested):

MNDonut (approved):

PR #1163, opened 2024-07-27
Solution

mihavryliuk:


app/main.py
```diff
@@ -1 +1,10 @@
-#  write your code here
+from typing import List, Union
+
+from app.animals.animal import Animal
+from app.animals.cat import Cat
+from app.animals.dog import Dog
+
+
+def feed_animals(all_animals: List[Union["Animal", "Cat", "Dog"]]) -> int:
```
LiudmylaKulinchenko, line 8: What if you'll keep only Animal in your annotations, will it work?

app/main.py
```diff
@@ -1 +1,10 @@
...
+    all_food = sum(animal.feed() for animal in all_animals)
+    return all_food
```
LiudmylaKulinchenko, line 10: You can put it directly in return to avoid creating a new variable that's used only once

LiudmylaKulinchenko (changes requested):

app/animals/cat.py
```diff
@@ -0,0 +1,11 @@
+from app.animals.animal import Animal
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
```
MNDonut, line 8: no need to pass required parameters explicitly. we often do it when we're gonna pass not required parameter and there are many of them like
```
def func(param1, param2, param3, unrequired1=None, unrequired2=None, unrequired3=None, unrequired4=None, unrequired5=None, unrequired6=None):
    ...

# call function with explicit unrequired6 param so that Python knows that "Ihor" isn't a value for unrequired1
func(1, 2, 3, unrequired6="Ihor")
# instead of this
func(1, 2, 3, None, None, None, None, None, "Ihor")
```
Fix everywhere

MNDonut (changes requested):

LiudmylaKulinchenko (changes requested):
You didn't fix Ihor's comment everywhere and why do you have your classes duplicated?

oliasymak24 (approved):

PR #1164, opened 2024-07-28
Solution

mnezbrytska:


app/main.py
```diff
@@ -1 +1,50 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            self.is_hungry = False
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        else:
+            return 0
```
LiudmylaKulinchenko, line 20: No need to use `else` if you have `return` in your `if` statement

app/main.py
```diff
@@ -1 +1,50 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.name = name
+        self.is_hungry = is_hungry
+        Dog.appetite = 7
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
LiudmylaKulinchenko, line 45: Let's also annotate what objects we expect to have in list

app/main.py
```diff
@@ -1 +1,50 @@
...
+    count_food = 0
+    for animal in animals:
+        if animal.is_hungry:
+            count_food += animal.feed()
+    return count_food
```
LiudmylaKulinchenko, line 50: U can rewrite it in one line using sum

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,45 @@
-#  write your code here
+class Animal:
+
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
```
MNDonut, line 6: you know what to do ;)

app/main.py
```diff
@@ -1 +1,45 @@
...
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        self.name = name
+        self.is_hungry = is_hungry
+        Cat.appetite = 3
```
MNDonut, line 27: You don't need to do it manually because you inherit from Animal, whose init makes the same, thus you can call super init. Fix everywhere

MNDonut (changes requested):

oliasymak24 (approved):
Good job!

PR #1172, opened 2024-07-28
Resolve

ihorutkin:


app/main.py
```diff
@@ -1 +1,56 @@
...
+
+
+def feed_animals(list_of_animals: list) -> int:
+    hungry_animals = []
+    for animal in list_of_animals:
+        if animal.is_hungry:
+            animal.feed()
+            hungry_animals.append(animal.appetite)
+
+    return sum(hungry_animals)
```
LiudmylaKulinchenko, line 56: You can rewrite it in one line

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,54 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
```
MNDonut, line 29: no need to pass required parameters explicitly. we often do it when we're gonna pass not required parameter and there are many of them like
```
def func(param1, param2, param3, unrequired1=None, unrequired2=None, unrequired3=None, unrequired4=None, unrequired5=None, unrequired6=None):
    ...

# call function with explicit unrequired6 param so that Python knows that "Ihor" isn't a value for unrequired1
func(1, 2, 3, unrequired6="Ihor")
# instead of this
func(1, 2, 3, None, None, None, None, None, "Ihor")
```
Fix everywhere

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,54 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    return sum([
+        animal.feed()
+        for animal in list_of_animals
+        if animal.is_hungry
```
MNDonut, line 53: no need to check if hungry because there is that condition in .feed()

MNDonut (changes requested):

app/main.py
MNDonut: You still pass parameters explicitly

app/main.py
```diff
@@ -1 +1,53 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=7)
```
MNDonut, line 42: still not fixed

app/main.py
```diff
@@ -1 +1,53 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    return sum([
+        animal.feed()
+        for animal in list_of_animals
+    ])
```
MNDonut, line 53: you can fit it in 1 line

MNDonut (changes requested):

MNDonut (changes requested):
You don't need to pass parameters using parameter's name if you follow the ordering that you set in init, just pass values

ihorutkin:
> You don't need to pass parameters using parameter's name if you follow the ordering that you set in init, just pass values

Yeah, now, during Q&A I understand it

app/main.py
```diff
@@ -1 +1,50 @@
...
+    ) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
+    return sum([animal.feed() for animal in list_of_animals])
```
oliasymak24, line 50: `sum` can work with generators, so [] are not needed

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1175, opened 2024-07-29
add 3 class(Animal, Dog and Cat)

egor250330:


app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
MNDonut, line 20: You don't need else statement because there is return in if

app/main.py
```diff
@@ -1 +1,45 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
```
MNDonut, line 34: no need to pass required params explicitly, so you can remove parameters' name

app/main.py
```diff
@@ -1 +1,45 @@
...
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list) -> int:
+    total_food_points = 0
+    for animal in animal_list:
+        food_points = animal.feed()
+        total_food_points += food_points
+    return total_food_points
```
MNDonut, line 45: you can do it in 1 line

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌸

PR #1176, opened 2024-07-29
Solution

olena4reunited:


app/main.py
```diff
@@ -1 +1,46 @@
...
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry is False:
```
oliasymak24, line 16: `not self.is_hungry ` is equal to `self.is_hungry is False`

app/main.py
```diff
@@ -1 +1,46 @@
...
+            return 0
+        print(F"Eating {self.appetite} food points...")
+        self.is_hungry = False
+
+        return self.appetite
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=3, is_hungry=is_hungry)
```
oliasymak24, line 26: You don't need  to pass required parameters explicitly, only for parameter which was changed will be better in this case
Check in all places

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
```
oliasymak24, line 36: Why do you need here self parameter?
Can just use staticmethod?

app/main.py
```diff
@@ -1 +1,46 @@
...
+        self.is_hungry = False
+
+        return self.appetite
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
```
oliasymak24, line 28: Why do you need here self parameter?
Can just use staticmethod?

app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_food_points = 0
+    for animal in animals:
+        food_points = animal.feed()
+        total_food_points += food_points
```
oliasymak24, line 44: Use sum() function instead)

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,45 @@
...
+        print(F"Eating {self.appetite} food points...")
+        self.is_hungry = False
+
+        return self.appetite
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        appetite = 3
+        super().__init__(name=name, appetite=appetite, is_hungry=is_hungry)
```
oliasymak24, line 27: I meant that it was not a mistake, just no need to do this,
just you can do (name, 3, is_hungry)

olena4reunited: Oh, i get it. Thank you!

PR #1179, opened 2024-07-29
task done!

Sergi0bbb:


app/main.py
```diff
@@ -1 +1,46 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: List[Animal]) -> int:
```
LiudmylaKulinchenko, line 45: You don't have to import List from typing, you can use `list` in the newest python versions

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 👀

PR #1217, opened 2024-09-02
Solution_1

apmaz:


app/main.py
```diff
@@ -1 +1,68 @@
-#  write your code here
+class Animal:
+    list_animals = []
+
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True
+                 ) -> None:
```
Dimosphen1, line 8: Place each argument including `self` on a new line here and in the other classes so that there would be no arguments on the same line with opening and closing parenthesis. It will help to improve the code style and readability

app/main.py
```diff
@@ -1 +1,68 @@
-#  write your code here
+class Animal:
+    list_animals = []
```
Dimosphen1, line 2: There is no need to indicate type in the attribute name, remove it
```suggestion
    animals = []
```

app/main.py
```diff
@@ -1 +1,68 @@
...
+            return self.appetite
+        print(f"Eating {self.appetite} food points...")
+        self.is_hungry = False
+        return self.appetite
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 appetite: int = 3,
```
Dimosphen1, line 31: Do not add `appetite` in the `__init__` for `Cat` and `Dog` classes, remove them as mentioned in the task description and pass required value directly to the `super().__init__`

app/main.py
```diff
@@ -1 +1,68 @@
...
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal: list) -> int:
```
Dimosphen1, line 64: Indicate that you expect to obtain a list of `Animal` class instances for `animals` in the type annotation. Argument name should be `animals` as well

app/main.py
```diff
@@ -1 +1,68 @@
...
+    food_count = 0
+    for food in animal:
+        food_count += Animal.feed(food)
+    return food_count
```
Dimosphen1, line 68: There is no food that is stored in the `animals`, there are `Animal` class instances. Call `feed` method from them to feed. Besides that, you could implement this method in one line using `sum`

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,62 @@
...
+        return self.appetite
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            appetite: int = 3,
+            is_hungry: bool = True
+    ) -> None:
```
pavlopro, line 35: Do not add appetite in the __init__ for Cat and Dog classes, remove them as mentioned in the task description and pass required value directly to the super().__init__

pavlopro (changes requested):

PR #1222, opened 2024-09-11
'Solution'

KrenkySpenky:


app/main.py
```diff
@@ -1 +1,54 @@
...
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True
+                 ) -> None:
```
Dimosphen1, line 28: Place each argument including `self` on a new line so that there would be no arguments on the same line with opening and closing parenthesis here and in the `Dog` class as well. It will help to improve the code style and readability

app/main.py
```diff
@@ -1 +1,54 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> float:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
Dimosphen1, line 20: There is no need for `else` since the code will not execute after the `return` statement anyway. Please, remove the redundant `else`

app/main.py
```diff
@@ -1 +1,54 @@
-#  write your code here
+class Animal:
+    def __init__(
+            self,
+            name: str,
+            appetite: float,
```
Dimosphen1, line 5: `appetite` could be `int` as well, add it to the type annotation

app/main.py
```diff
@@ -1 +1,54 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_food_points = 0
+    for animal in animals:
+        if animal.is_hungry:
+            total_food_points += animal.feed()
+    return total_food_points
```
Dimosphen1, line 54: Implement this method in one line using `sum`. There is no need to check if animal is hungry here since it is done in the `feed` method

Dimosphen1 (changes requested):

app/main.py
KrenkySpenky: image

app/main.py
```diff
@@ -1 +1,51 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> float:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
Dimosphen1, line 21: Please, remove the redundant `else` and leave the return statement
```suggestion
        return 0
```

app/main.py
```diff
@@ -1 +1,51 @@
...
+    ) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum(animal.feed() for animal in animals if animal.is_hungry)
```
Dimosphen1, line 51: There is no need to perform the if check here, it is already done in the `feed` method
```suggestion
    return sum(animal.feed() for animal in animals)
```

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #1226, opened 2024-09-16
Solution

Viacheslav417865:


app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = 0
+    for animal in animals:
```
vasylhnatiuk, line 38: use list comprehension

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
vasylhnatiuk, line 36:
```suggestion
def feed_animals(animals: list[Animal]) -> int:
```

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
```
vasylhnatiuk, line 16: else is redundnat here

vasylhnatiuk (changes requested):

app/main.py
YukihiroSM: Still not fixed. You can use list comprehension here as an improvement.

YukihiroSM (changes requested):

PR #1229, opened 2024-09-21
feed animals

dmytrik:


app/main.py
```diff
@@ -1 +1,50 @@
-#  write your code here
+class Animal:
+
+    def __init__(self, name: str, appetite: int,
+                 is_hungry: bool = True) -> None:
```
AnyoneClown, line 4: Bad identation, try to locate every parametr on a new line

app/main.py
```diff
@@ -1 +1,50 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    sum_food_points = 0
+
+    for animal in animals:
+        sum_food_points += animal.feed()
+
+    return sum_food_points
```
AnyoneClown, line 50: Let's make return with one-liner. Use `sum` for that

AnyoneClown (changes requested):

PR #1259, opened 2024-10-06
Solution

Rabb1t-Artem:


VolodymyrSemchysyn (approved):

app/main.py
```diff
@@ -1 +1,40 @@
...
+            return self.appetite
+        else:
+            return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    def catch_mouse(self) -> str:
```
VolodymyrSemchysyn, line 25: i think it's not necessarily to return "The hunt began!" because you already print it before and you can just do annotition type -> None

app/main.py
```diff
@@ -1 +1,40 @@
...
+        print("The hunt began!")
+        self.is_hungry = False
+        return "The hunt began!"
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
```
VolodymyrSemchysyn, line 35: and this method should be static coz you don't need "self" to implement it

app/main.py
```diff
@@ -1 +1,40 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            self.is_hungry = False
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        else:
+            return 0
```
Nattalli, line 18: you don'tneed `else` statement

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    def catch_mouse(self) -> str:
+        print("The hunt began!")
+        self.is_hungry = False
+        return "The hunt began!"
```
Nattalli, line 28: that is redundant, remove it

Nattalli: that ethod should print only

Nattalli (changes requested):

PR #1273, opened 2024-10-08
Solution

vmyronets:


app/main.py
```diff
@@ -1 +1,42 @@
...
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_of_animals: list) -> int:
```
Dimosphen1, line 41: There is no need to indicate data type in a variable name without a need. Rename `list_of_animals` to `animals`. Besides that, indicate that you expect to obtain a list of `Animal` class instances for `animals` in the type annotation

Dimosphen1 (changes requested):

app/main.py
vmyronets: Ok, thanks

PR #1291, opened 2024-10-20
solution

popeye88:


animal/animal.py
```diff
@@ -0,0 +1,17 @@
+class Animal:
```
pavlopro, line 1: There is no need to have this class defined in separate file. Move this class definition to main.py.

app/main.py
```diff
@@ -1 +1,28 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> None:
+    food_points = 0
```
pavlopro, line 23: Try to solve it in one line

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #1327, opened 2024-11-16
Solution

bodiakof:


app/main.py
```diff
@@ -1 +1,46 @@
...
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 7) -> None:
```
L1nk3rrr: Re-read task requirements. It should have the init method with two arguments - name and is hungry

app/main.py
```diff
@@ -1 +1,46 @@
...
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True,
+            appetite: int = 3) -> None:
```
L1nk3rrr: Re-read task requirements. It should have the __init__ method with two arguments - name and is hungry

L1nk3rrr (changes requested):

.flake8
```diff
@@ -4,4 +4,4 @@ ignore = E203, E266, W503, ANN002, ANN003, ANN101, ANN102, ANN401, N807, N818
 max-line-length = 79
 max-complexity = 18
 select = B,C,E,F,W,T4,B9,ANN,Q0,N8,VNE
-exclude = venv, tests
+exclude = .venv, tests
```
L1nk3rrr: No need to commit such changes. If you change .flake8 do it locally, don't push such things

app/main.py
```diff
@@ -1 +1,44 @@
...
+            self,
+            name: str,
+            is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
```
L1nk3rrr, line 43: Looks like we accept only animals at list, so try to add annotation like `list[TYPE_NAME]`

L1nk3rrr (changes requested):
Return .flake8 to the previous state :)  Simply remove the dot before `venv` and push this.
From the task side all is good, well done!

L1nk3rrr (approved):
Well done!

PR #1328, opened 2024-11-17
solution

MaksymStarodubtsev:


app/main.py
```diff
@@ -1 +1,47 @@
...
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+
+            return self.appetite
+        else:
+            return 0
```
Nattalli, line 22: you can avoid `else` statement

Nattalli, line 22:
```suggestion
        if self.is_hungry:
            print(f"Eating {self.appetite} food points...")
            self.is_hungry = False

            return self.appetite
        return 0
```

app/main.py
```diff
@@ -1 +1,47 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: [Animal]) -> int:
+    count = 0
+
+    for animal in animals:
+        count += animal.feed()
+
+    return count
```
Nattalli, line 47: can you use list comprehension?

Nattalli: and `sum` method for count calculating

Nattalli (changes requested):

PR #1337, opened 2024-11-24
Created all classes and methods

shafran-andriy:


app/main.py
```diff
@@ -1 +1,50 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str,
+                 is_hungry: bool = True,
+                 appetite: int = 3,
```
pavlopro: Please remove default value for appetite. Instead pass it into `super`

app/main.py
```diff
@@ -1 +1,50 @@
...
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str,
+                 is_hungry: bool = True,
+                 appetite: int = 7,
+                 ) -> None:
+        super().__init__(name, appetite, is_hungry)
```
pavlopro: Please remove default value for appetite. Instead pass it into `super`

pavlopro (changes requested):

app/main.py
shafran-andriy: done

app/main.py
shafran-andriy: done

PR #1350, opened 2024-11-29
Solution

julia4406:


app/main.py
```diff
@@ -1 +1,44 @@
...
+                 is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    return sum([beast.feed() for beast in animals])
```
L1nk3rrr: You can simlify here
` sum(beast.feed() for beast in animals)`

L1nk3rrr (changes requested):
Well done!
Left some minor styling points, also
`    def __init__(self,
                 name: str,
                 appetite: int,
                 is_hungry: bool = True) -> None:`
change to
`    def __init__(self,
                 name: str,
                 appetite: int,
                 is_hungry: bool = True
        ) -> None:`
        
 It is better styling practice

app/main.py
julia4406: Hi. 
Thanks for hint with iterator, got it.
But I'm a little puzzled with the comment above. Did you mean better code should be in one line?

L1nk3rrr: Yep, idk why github refused my styling...

L1nk3rrr (approved):
Well done!

PR #1365, opened 2024-12-05
Solution

bogdAAAn1:


app/main.py
```diff
@@ -1 +1,51 @@
-#  write your code here
+class Animal:
+    def __init__(self,
+                 name: str,
+                 appetite: int,
+                 is_hungry: bool = True
+                 ) -> None:
```
L1nk3rrr: Try to do one attribute per line. so it will looks like
```
    def __init__(
         self,
         name: str,
         appetite: int,
         is_hungry: bool = True
    ) -> None:
```
And `) -> None` always below of method/function name started

app/main.py
```diff
@@ -1 +1,51 @@
...
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True
+                 ) -> None:
```
L1nk3rrr: Same styling as on Animal

app/main.py
```diff
@@ -1 +1,51 @@
...
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self,
+                 name: str,
+                 is_hungry: bool = True
+                 ) -> None:
```
L1nk3rrr: Same styling as on Animal

L1nk3rrr (changes requested):
Well done! Some styling improvements for you, please change it

L1nk3rrr (approved):
Nice! Great Work!

PR #1367, opened 2024-12-06
'Solution'

Dogober:


app/main.py
```diff
@@ -1 +1,52 @@
-#  write your code here
+class Animal:
+
+    def __init__(
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True) -> None:
```
julia4406: if you're using "(" on separate row, you should use "): -> None" on separate row too.

app/main.py
```diff
@@ -1 +1,52 @@
...
+
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+
```
julia4406: blank line not need here, it complicates readability

app/main.py
```diff
@@ -1 +1,52 @@
...
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+
```
julia4406: the same with black line - it's hard to read what method do

julia4406 (approved):
Cool work! I added some minor comments, just for improving.

app/main.py
```diff
@@ -1 +1,49 @@
-#  write your code here
+class Animal:
+
+    def __init__(
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True) \
+            -> None:
```
L1nk3rrr: From the styling side, at one of your checklist, you had an example like this:
```
    def __init__(
            self,
            name: str,
            appetite: int,
            is_hungry: bool = True
    )-> None:
```

app/main.py
```diff
@@ -1 +1,49 @@
...
+        super().__init__(name, appetite=7)
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    return sum([animal.feed() for animal in animals])
```
L1nk3rrr: You don't need a list here, cause you will have an iterator, look at checklist

L1nk3rrr (changes requested):
Well done! Minor change required

L1nk3rrr (approved):
Well done!

PR #1377, opened 2024-12-08
Solution

IvankaKuzin:


app/main.py
```diff
@@ -1 +1,59 @@
...
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=3)
+        self.name = name
+        self.is_hungry = is_hungry
```
TretiakHelen: Since the parent class Animal already includes the "name" and "is_hungry" attributes, we don't need to redefine it in the Cat class.

app/main.py
```diff
@@ -1 +1,59 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=3)
```
TretiakHelen: We can just use "super().__init__(name, 3, is_hungry)"

app/main.py
```diff
@@ -1 +1,59 @@
...
+
+class Dog(Animal):
+    def __init__(
+            self,
+            name: str,
+            is_hungry: bool = True
+    ) -> None:
+        super().__init__(name, appetite=7)
+        self.name = name
+        self.is_hungry = is_hungry
```
TretiakHelen: Same situation as in the Cat class

app/main.py
```diff
@@ -1 +1,59 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    sum_of_food_points = 0
+
+    for animal in animals:
+        if animal.is_hungry:
+            sum_of_food_points += animal.appetite
+            animal.feed()
```
TretiakHelen: Since animal.feed() already handles changing is_hungry to False and returning the appropriate value, you can call feed() directly and accumulate its return value.
```
for animal in animals:
        sum_of_food_points +=  animal.feed()
```

TretiakHelen (approved):
It works properly and has passed all checks.

PR #1382, opened 2024-12-09
solved

Danylo-Ok:


10kkyvl (approved):

app/main.py
```diff
@@ -1 +1,40 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str,
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
+        self.name = name
```
petrykivd, line 5: Each parameter must be from a new line

app/main.py
```diff
@@ -1 +1,40 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food = 0
+    for animal in animals:
+        total_food += animal.feed()
+    return total_food
```
petrykivd, line 40: Please read checklist before pushing your solution. 
Point 2 in checklist: Don't use a for loop if you can use generator expression.

petrykivd (changes requested):

PR #1385, opened 2024-12-09
Solution

Bleit1:


Sichkaridze (approved):

app/main.py
```diff
@@ -1 +1,46 @@
-#  write your code here
+
+class Animal:
+    def __init__(self, name: str,
+                 appetite: int,
+                 is_hungry : bool = True) -> None:
```
petrykivd, line 5: Fix codestyle, every parameter must be from a new line

petrykivd: In all places

app/main.py
```diff
@@ -1 +1,46 @@
...
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals : list) -> None:
+    total_food = 0
+    for animal in animals:
+        eaten_food = animal.feed()
+        total_food += eaten_food
+    return total_food
```
petrykivd, line 46: You can add a type of objects in list.
Also, please read a checklist before pushing your solution.
Point 2 in checklist: Don't use a for loop if you can use generator expression.

petrykivd (changes requested):

PR #1392, opened 2024-12-09
Solution

toku666:


app/main.py
```diff
@@ -1 +1,39 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str,
+                 appetite: int, is_hungry: bool = True) -> None:
```
Professor-Douel: change format like:
(self,
 name: str,
 appetite: int,
 is_hungry: bool = True) -> None:

app/main.py
```diff
@@ -1 +1,39 @@
...
+    def feed(self) -> int:
+        if self.is_hungry:
+            self.is_hungry = False
+            print(f"Eating {self.appetite} food points...")
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
```
Professor-Douel, line 22: here too:
(self,
 name: str,
 is_hungry: bool = True) -> None:

app/main.py
```diff
@@ -1 +1,39 @@
...
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
```
Professor-Douel, line 30: here too:
(self,
 name: str,
 is_hungry: bool = True) -> None:

Professor-Douel (changes requested):

Professor-Douel (changes requested):
You need to change code formatting in two more places

PR #1409, opened 2024-12-17
all tests passed. flake8 app passed

Mykola-Osolinskyi:


app/main.py
```diff
@@ -1 +1,45 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list["Animal"]) -> int:
+    return sum(Animal.feed(animal) for animal in animals)
```
Nattalli: that would be better:

```suggestion
    return sum(animal.feed() for animal in animals)
```

Nattalli (changes requested):

app/main.py
Mykola-Osolinskyi: @Nattalli 
refactored according to your recommendations

Nattalli (approved):

PR #1421, opened 2024-12-29
implement Animal, Cat and Dog classes. implement feed_animals function

yarzhuravlov:


app/main.py
```diff
@@ -1 +1,37 @@
...
+
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
```
oliasymak24: No need for self, make like static method

app/main.py
```diff
@@ -1 +1,37 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
```
oliasymak24: Same static method

oliasymak24 (changes requested):

app/main.py
yarzhuravlov: Now the method is static, thanks for the feedback.

app/main.py
yarzhuravlov: Now the method is static, thanks for the feedback.

oliasymak24 (approved):

PR #1426, opened 2025-01-05
Solution

Baranotik15:


app/main.py
```diff
@@ -1 +1,45 @@
...
+
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, is_hungry=is_hungry, appetite=3)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
```
oliasymak24: No need for self, can be static

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, is_hungry=is_hungry, appetite=7)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
```
oliasymak24: No need for self, can be static

app/main.py
```diff
@@ -1 +1,45 @@
...
+
+
+def feed_animals(list_of_animals: list[Animal]) -> int:
+    result = 0
+    for animal in list_of_animals:
+        result += animal.feed()
+    return result
```
oliasymak24: use function sum

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1464, opened 2025-01-28
Виконав завдання

GGZahar1:


app/main.py
```diff
@@ -1 +1,51 @@
...
+        super().__init__(name,
+                         appetite=7,
+                         is_hungry=is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(list_animals: list) -> None:
```
vsmutok: wrong return type annotation

app/main.py
```diff
@@ -1 +1,51 @@
...
+    count = 0
+    for animals in list_animals:
+        if animals.is_hungry:
+            count += animals.appetite
+            animals.feed()
+    return count
```
vsmutok: Try not to use a loop in this task; instead, you can use `sum` and a generator.

vsmutok (changes requested):

PR #1485, opened 2025-02-14
Finish

Sotnikov-100:


app/main.py
```diff
@@ -1 +1,42 @@
-#  write your code here
+class Animal:
+    def __init__(
+            self, name: str, appetite: int, is_hungry: bool = True
```
Baranotik15: image-removebg-preview
Correctly move each next attribute to a new line

Baranotik15 (changes requested):
Good Job but need some change

app/main.py
Sotnikov-100: Thank you Bro, I forgot That

Baranotik15 (approved):
Good job!

PR #1501, opened 2025-02-22
Solution

HappyDen08:


AndyStarGitHub (approved):
Hello! Well done! I only wonder if two specific methods in Cat and Dog classes should be static. But if it works, then it is fine.

PR #1519, opened 2025-02-24
Feed animals (1.0)

Maaax256:


app/main.py
```diff
@@ -1 +1,41 @@
...
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    count_points = 0
+    for animal in animals:
```
Roksolana-K: Following the checklist: don't use a for loop if you can use generator expression.

Roksolana-K (changes requested):
Nice job! However, I have a few suggestions. Please, give a look.

app/main.py
Maaax256: fair enough

PR #1620, opened 2025-04-30
Solution

Nib20:


app/main.py
```diff
@@ -1 +1,46 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    result_sum = 0
+    for animal in animals:
+        result_sum += animal.feed()
+    return result_sum
```
MNDonut: let's use built-it **sum** function here ;)

MNDonut (changes requested):

app/main.py
Nib20: fixed

MNDonut (approved):
awesome!

PR #1688, opened 2025-06-22
Solution

zqpiq:


app/main.py
```diff
@@ -1 +1,43 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
MNDonut: you can remove `else` because if statement has `return`

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    counter_food = 0
+    for animal in animals:
+        counter_food += animal.feed()
+    return counter_food
```
MNDonut: you can use `sum` function here

MNDonut (changes requested):

MNDonut (approved):

PR #1692, opened 2025-06-25
Solution for py-feed-animals

Hetmarum:


app/main.py
```diff
@@ -1 +1,45 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    points = 0
+
+    for animal in animals:
+        points += animal.feed()
```
Arch0998, line 44: Don't use a `for` loop if you can use generator expression.

**Good example:**
```
return sum(item.process() for item in collection)
```

Bad example:
```
total = 0
for item in collection:
    total += item.process()
return total
```

app/main.py
```diff
@@ -1 +1,45 @@
-#  write your code here
+class Animal:
+    def __init__(
+            self,
+            name: str,
+            appetite: int,
+            is_hungry: bool = True
+    ) -> None:
+
```
Arch0998, line 8: It is not necessary to add an extra blank line between the closing parenthesis and return annotation () -> None:) and the first line of the method body (self.name = name).
The current formatting, where the method body starts immediately after the method signature with proper indentation, is clear and follows common Python style conventions (PEP 8).
**Adding a blank line there is optional but generally not recommended**, as it can reduce code compactness and readability. Keeping the method body directly after the signature makes the code cleaner and easier to follow.

app/main.py
```diff
@@ -1 +1,45 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, is_hungry=is_hungry, appetite=7)
```
Arch0998, line 34: If you want to be as tidy as possible, you can keep the keyword arguments in the same order for better readability, but this is a matter of style, not an error.
For example, in the previous super() call the arguments are ordered like this:

`super().__init__(name, appetite=3, is_hungry=is_hungry)
`
while in this super() call they are ordered like this:
`
super().__init__(name, is_hungry=is_hungry, appetite=7)
`

Arch0998 (approved):
Everything works correctly, but please make sure to read the file **checklist.md**, where described the recommended code style guidelines to follow.

PR #1703, opened 2025-06-26
Solution

Artcrafterrra:


app/main.py
```diff
@@ -1 +1,45 @@
...
+        is_hungry: bool = True
+    ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int | None:
```
dimashevchukk, line 15: can't return None

app/main.py
```diff
@@ -1 +1,45 @@
...
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total = 0
+    for animal in animals:
+        total += animal.feed()
+    return total
```
dimashevchukk, line 45: can use generator comprehension with sum() instead

dimashevchukk (approved):

PR #1709, opened 2025-06-27
Implemented solution.

Otst0y:


app/main.py
```diff
@@ -1 +1,55 @@
...
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
```
VadymKyrylko: You should place each argument on a new line (see checklist.md)

app/main.py
```diff
@@ -1 +1,55 @@
...
+        super().__init__(name, appetite=3, is_hungry=is_hungry)
```
VadymKyrylko: It would be better if you implement some changes here.
Remove the "appetite=" and remove the next line.

app/main.py
```diff
@@ -1 +1,55 @@
...
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
```
VadymKyrylko: Same:
You should place each argument on a new line (see checklist.md)

app/main.py
```diff
@@ -1 +1,55 @@
...
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
```
VadymKyrylko: Same: you don't need to use "appetite=", only 7.
And you should remove the next line.

app/main.py
```diff
@@ -1 +1,55 @@
...
+        self.is_hungry = is_hungry
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> None:
+    appetites = 0
```
VadymKyrylko: The condition of the problem says that we have to count the total amount of food we give to the animals. It is advisable to use the name of the variable in relation to food, not the appetites of the animals.

app/main.py
```diff
@@ -1 +1,55 @@
...
+
+    for animal in animals:
+        if animal.is_hungry:
+            animal.feed()
```
VadymKyrylko: This method can correctly works in the next line. Change "animal.appetite" to "animal.feed()" in the next line, and remove this line.

VadymKyrylko (changes requested):
Not bad! But look in the comments

PR #1728, opened 2025-07-01
Solution

Irahromska:


app/main.py
```diff
@@ -1 +1,46 @@
...
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, appetite=7, is_hungry=is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: List[Animal]) -> int:
```
bohdankolisnyk04, line 42: from checklist:

2. Don't use a `for` loop if you can use generator expression.

Good example:

```python
return sum(item.process() for item in collection)
```

Bad example:

```python
total = 0
for item in collection:
    total += item.process()
return total
```

bohdankolisnyk04 (changes requested):

MNDonut (approved):

PR #1820, opened 2025-09-23
Solution

3883IZ:


app/main.py
```diff
@@ -1 +1,53 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: List[Animal]) -> int:
+    total_food = 0
+    for animal in animals:
+        total_food += animal.feed()
+    return total_food
```
vsmutok: This implementation works, but it violates a guideline from the project's checklist. According to checklist item https://github.com/mate-academy/py-feed-animals/pull/2 under 'Make Code Easier': "Don't use a `for` loop if you can use generator expression."

You should use a generator expression with the `sum()` function instead of a `for` loop to calculate the total. This approach is more concise and Pythonic.

Could you refactor this to use `sum()` and a generator expression?

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1861, opened 2025-10-27
add solution

VadymVynokur:


app/main.py
```diff
@@ -1 +1,41 @@
...
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
```
Vorongor, line 29: Do you know if we need a reference to self here? It may be a static method.

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
```
Vorongor, line 37: Do we need a reference to self here? It may be a static method.

app/main.py
```diff
@@ -1 +1,41 @@
...
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
```
Vorongor, line 26: Do we have to set the appetite if Cat isn`t hungry? Could it be better to set it to 0

app/main.py
```diff
@@ -1 +1,41 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
```
Vorongor, line 34: Do we have to set the appetite if Dog isn`t hungry? Could it be better to set it to 0

Vorongor (approved):
Well done! Good structured code,  laconic and very clear, possibly some docstring may improve youre code.   Some Inaccuracies: set appetite if animal isn`t hungry and use staticmethod if no needed of self reference.

PR #1867, opened 2025-10-29
Solution

leemonch1k-space:


app/main.py
```diff
@@ -1 +1,62 @@
-#  write your code here
+from __future__ import annotations
```
stanislav-kishechko: No import is needed, as we do not have a case where a class references itself.

app/main.py
```diff
@@ -1 +1,62 @@
...
+
+
+class Animal:
+
```
stanislav-kishechko: Empty lines after class declarations are not necessary.

stanislav-kishechko (approved):

app/main.py
leemonch1k-space: Done

app/main.py
leemonch1k-space: will know

PR #1868, opened 2025-10-29
Solution

stanislav-kishechko:


VadymVynokur (approved):
so many docstrings 0_o, are u really wrote it by yourself? :)

PR #1872, opened 2025-10-30
Solution

tradewcs:


app/main.py
```diff
@@ -1 +1,44 @@
...
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if not self.is_hungry:
+            return 0
+
+        print(f"Eating {self.appetite} food points...")
+
+        tmp = self.appetite
+        self.appetite = 0
```
vakt159: you don`t need to set self.appetite to 0

vakt159 (changes requested):
remove self.appetite = 0

vakt159 (approved):

PR #1873, opened 2025-10-30
Solution

vakt159:


app/main.py
```diff
@@ -1 +1,42 @@
...
+                 appetite: int,
+                 is_hungry: bool = True) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int | None:
```
tradewcs: hint annotation is wrong, feed() is always returns int

tradewcs (changes requested):

PR #1874, opened 2025-10-30
Develop

AngryAndrii:


app/main.py
```diff
@@ -1 +1,48 @@
-#  write your code here
+from __future__ import annotations
```
Taras-Ivashchuk, line 1: Andrii,  this import is redundant in this case

Taras-Ivashchuk (approved):

PR #1883, opened 2025-10-31
Solution

tigersaymon:


app/main.py
```diff
@@ -1 +1,44 @@
...
+            self.appetite = 0
+            return eaten_amount
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=3)
+        self.name = name
+        self.is_hungry = is_hungry
```
n-korobko: Can be removed from the Cat class, because you're already calling super().__init__(),
which initializes these attributes in the parent (Animal) class.  self.name = name
        self.is_hungry = is_hungry  DELETED IT)

app/main.py
```diff
@@ -1 +1,44 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=7)
+        self.name = name
```
n-korobko: HERE ALSO DELETE self.name = name
        self.is_hungry = is_hungry

app/main.py
tigersaymon: I agree that self.name = name is unnecessary here. However, I received one failed test without self.is_hungry = is_hungry

n-korobko: Interesting)) Okey))

n-korobko (approved):

app/main.py
tigersaymon: This line could be deleted, it just needed to be added to the init method. Thank you

PR #1884, opened 2025-10-31
Feed animals

n-korobko:


VYGAKOOL (approved):
It works, but "catch-mouse" and "bring_slippers" may be static

PR #1889, opened 2025-11-01
created classes animal, cat, dog. and func feed_animals

AlexMishKo:


oleksii-shcherbak-code (approved):
Good work, i approve it, but, in my opinion, it would be more pythonic to make the catch_mouse and bring_slippers methods statics

PR #1894, opened 2025-11-03
solution

Akcolotle:


app/main.py
```diff
@@ -1 +1,40 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_food = 0
+    for animal in animals:
```
CookieTanuki: Use generator expression with sum() function instead of for loop

CookieTanuki (approved):

PR #1978, opened 2025-12-30
Solution

kovalevv:


kovalevv:
@mateacademy-ai-mentor please check the solution with the fix

PR #2016, opened 2026-01-22
Solution

viniciusrivello:


viniciusrivello (commented):
aprovaaa

viniciusrivello (commented):
aprova..

viniciusrivello (commented):
apro

viniciusrivello (commented):
from typing import List


class Animal:
    def __init__(self, name: str, appetite: int, is_hungry: bool = True) -> None:
        self.name = name
        self.appetite = appetite
        self.is_hungry = is_hungry

    def print_name(self) -> None:
        print(f"Hello, I'm {self.name}")

    def feed(self) -> int:
        if not self.is_hungry:
            return 0

        self.is_hungry = False
        return self.appetite


class Cat(Animal):

    def catch_mouse(self) -> None:
        print("The hunt began!")


class Dog(Animal):
        super().__init__(name, 7, is_hungry)

    def bring_slippers(self) -> None:
        print("The slippers delivered!")


def feed_animals(animals: List[Animal]) -> int:

PR #2030, opened 2026-01-29
Finish

Artem-byte123:
Finish task where I wrote code with inheritance classes from fathers class to child class


vsmutok (approved):
Great!

PR #2047, opened 2026-02-09
Solution: implementing function and classes to complete the task

SerhiiHryniuk:


app/main.py
```diff
@@ -1 +1,42 @@
-#  write your code here
+class Animal:
```
Reverad, line 4: u could add "from __future__ import annotations" to support older python versions.

app/main.py
```diff
@@ -1 +1,42 @@
...
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    @staticmethod
+    def catch_mouse() -> None:
+        print("The hunt began!")
```
Reverad: it would be better to use an instance method instead of a staticmethod. for example, this method could change attribute self.is_hungry in the future.

app/main.py
```diff
@@ -1 +1,42 @@
...
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
```
Reverad: same to previous comment.

Reverad (changes requested):

Reverad (approved):

PR #2098, opened 2026-03-04
Feed animals

V23IPPT:


Reverad (changes requested):
you made a PR from the main branch. for these tasks, we're supposed to create a separate feature branch first. you should move your changes to a new branch and revert the main one. it'll make the history much cleaner. 
it's always a good idea to read the guidelines first, btw. we use feature branches here to avoid messing up the main branch.

PR #2115, opened 2026-03-11
Solution

AlinaSv88:


app/main.py
```diff
@@ -1 +1,50 @@
...
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> None | int:
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        else:
+            return 0
```
vsmutok:
```suggestion
            return self.appetite
        return 0
```
`else`  redundant here

app/main.py
```diff
@@ -1 +1,50 @@
...
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    @staticmethod
+    def bring_slippers() -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> int:
+    total_food_points = 0
```
vsmutok: please try to write it in one line

vsmutok (changes requested):

app/main.py
AlinaSv88: Done

app/main.py
AlinaSv88: I rewrote this in one line, but now there are duplications. Is that what you meant, or should I try something different?

app/main.py
```diff
@@ -1 +1,52 @@
...
+def feed_animals(animals: list) -> int:
+    total_food_points = sum(
+        animal.appetite
+        for animal in animals
+        if animal.is_hungry
+    )
+    for animal in animals:
+        if animal.is_hungry:
+            animal.feed()
+    return total_food_points
```
vsmutok:
```suggestion
    return sum(animal.feed() for animal in animals)
```

vsmutok (changes requested):

vsmutok (approved):
Looks Good!

PR #2142, opened 2026-03-29
Solution

lubkoserhii:


app/main.py
```diff
@@ -1 +1,41 @@
...
+        if self.is_hungry:
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name=name, appetite=3, is_hungry=is_hungry)
```
PavelPereverzev1: Passing the appetite value this way is hardcoding; every cat will always have a value of 3, even if it's a tiger. Also, since the parent class parameters are positional, it's enough to simply maintain their order.
`class Cat(Animal):
    def __init__(
        self, 
        name: str, 
        appetite: int = 3, 
        is_hungry: bool = True
    ) -> None:
        super().__init__(name, appetite, is_hungry)`
I think this should be a better way.
The same in Dog class

PavelPereverzev1 (changes requested):

app/main.py
lubkoserhii: @PavelPereverzev1 
Done!

PR #2156, opened 2026-04-04
Solution

dimaaa-cmd:


app/main.py
```diff
@@ -1 +1,38 @@
-#  write your code here
+class Animal:
+    def __init__(self, name: str, appetite: int, is_hungry: bool = True) \
```
VadimPusiak788: Recommends avoiding backslash \ for line continuation. Instead, use implicit continuation inside parentheses ()

app/main.py
```diff
@@ -1 +1,38 @@
...
+            -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
```
VadimPusiak788: Use an early return for improved readability and reduced nesting

app/main.py
```diff
@@ -1 +1,38 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
```
VadimPusiak788: It would be better to use a default value in __init__ or define a class constant, since hardcoded values are not recommended.

VadimPusiak788 (changes requested):

app/main.py
dimaaa-cmd: I can move the constants outside the class or use default values. Because if I define a constant within the class scope, the tests don't work properly.

VadimPusiak788: DEFAULT_APPETITE = 7

        super().__init__(name, self.DEFAULT_APPETITE, is_hungry)
        
   OR
   
   
    def __init__(self, name: str, appetite: int =7, is_hungry: bool = True) -> None:
        super().__init__(name, appetite, is_hungry)

app/main.py
petrykivd: Listen to Vadim's comment

petrykivd (approved):
Nice! 🚀

VadimPusiak788 (approved):

PR #2157, opened 2026-04-04
Implement code

ToJaVadym:


app/main.py
```diff
@@ -1 +1,51 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list[Animal]) -> int:
+    total_points = 0
+    for animal in animals:
+        total_points += animal.feed()
+    return total_points
```
vsmutok: This violates checklist item [#2](https://github.com/mate-academy/py-feed-animals/pull/2): 'Don't use a `for` loop if you can use generator expression.' The logic can be simplified by using the built-in `sum()` function with a generator expression to calculate the total food points.

vsmutok (changes requested):

Boyok1337 (approved):
LGTM!

PR #2174, opened 2026-04-07
Solution

WachaX52:


app/main.py
```diff
@@ -1 +1,48 @@
...
+    ) -> None:
+        self.name: str = name
+        self.appetite: int = appetite
+        self.is_hungry: bool = is_hungry
+
+    def print_name(self) -> None:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> int:
+        if self.is_hungry:
```
VadimPusiak788: Use an early return for improved readability and reduced nesting

app/main.py
```diff
@@ -1 +1,48 @@
...
+            print(f"Eating {self.appetite} food points...")
+            self.is_hungry = False
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        # Appetite для котів фіксований — 3
+        super().__init__(name, 3, is_hungry)
```
VadimPusiak788: It would be better to use a default value in init o since hardcoded values are not recommended.

app/main.py
```diff
@@ -1 +1,48 @@
...
+
+    def catch_mouse(self) -> None:
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        # Appetite для собак фіксований — 7
+        super().__init__(name, 7, is_hungry)
```
VadimPusiak788: It would be better to use a default value in init o since hardcoded values are not recommended.

VadimPusiak788 (changes requested):

VadimPusiak788 (approved):

PR #2301, opened 2026-07-20
Solution

idlt044:


idlt044:
@mateacademy-ai-mentor I can't Remove the Animal inheritance from the Dog class because after this class dog won't have a functional feed, so it's correct

PR #2304, opened 2026-07-22
Solution

perehinets:


app/main.py
```diff
@@ -1 +1,38 @@
...
+                 is_hungry: bool = True
+                 ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> str:
+        print(f"Hello, I'm {self.name}")
+
+    def feed(self) -> str:
```
hladiivolodymyr: Since this function returns int, please update the type annotation accordingly.

app/main.py
```diff
@@ -1 +1,38 @@
...
+class Animal:
+    def __init__(self, name: str,
+                 appetite: int,
+                 is_hungry: bool = True
+                 ) -> None:
+        self.name = name
+        self.appetite = appetite
+        self.is_hungry = is_hungry
+
+    def print_name(self) -> str:
```
hladiivolodymyr: Since this function only prints a message and doesn't return anything, please change the return type annotation to None

app/main.py
```diff
@@ -1 +1,38 @@
...
+            self.is_hungry = not self.is_hungry
+            return self.appetite
+        return 0
+
+
+class Cat(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 3, is_hungry)
+
+    def catch_mouse(self) -> str:
```
hladiivolodymyr: Since this function only prints a message and doesn't return anything, please change the return type annotation to None

app/main.py
```diff
@@ -1 +1,38 @@
...
+        print("The hunt began!")
+
+
+class Dog(Animal):
+    def __init__(self, name: str, is_hungry: bool = True) -> None:
+        super().__init__(name, 7, is_hungry)
+
+    def bring_slippers(self) -> str:
```
hladiivolodymyr: Since this function only prints a message and doesn't return anything, please change the return type annotation to None

app/main.py
```diff
@@ -1 +1,38 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animals: list) -> None:
```
hladiivolodymyr: This function returns an integer, so please change the return type annotation to int

hladiivolodymyr (changes requested):

hladiivolodymyr (approved):

PR #2312, opened 2026-07-30
Sending the first solution

JessieMontaro:


app/main.py
```diff
@@ -1 +1,45 @@
...
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list) -> int:
+    hunger = 0
+    for animal in animal_list:
+        if animal.is_hungry:
+            hunger += animal.appetite
+            animal.feed()
+    return hunger
```
vsmutok: This violates **checklist item #2**: use a generator expression with `sum()` instead of a for loop with accumulation. The recommended pattern is: `return sum(animal.appetite for animal in animal_list if animal.is_hungry)` followed by feeding them separately, or restructure to accumulate and feed in a single pass using a generator.

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,43 @@
...
+
+    def bring_slippers(self) -> None:
+        print("The slippers delivered!")
+
+
+def feed_animals(animal_list: list) -> int:
+    hunger = sum(animal.appetite for animal in animal_list if animal.is_hungry)
+    for animal in animal_list:
+        animal.feed()
+    return hunger
```
vsmutok:
```suggestion
def feed_animals(animals: list) -> int:
    return sum(animal.feed() for animal in animals)
```

vsmutok (changes requested):

